PR #61, opened 2015-01-05
Fixes https://github.com/lightblue-platform/lightblue-core/issues/199

dcrissman:
Adds support to lightblue-rest to check for the new noone and anyone permissions.


dcrissman:
Depends on https://github.com/lightblue-platform/lightblue-core/pull/270

luan-cestari:
Tried twice and Travis claims some erros on the arquillian tests, maybe that known travis issue. Could you confirm that it is working locally? (so we can merge, as the master branch build usually doesnt have this kind of error)
https://travis-ci.org/lightblue-platform/lightblue-rest/builds/46000568

metadata/src/main/java/com/redhat/lightblue/rest/metadata/AbstractMetadataResource.java
```diff
@@ -340,18 +288,54 @@ public String removeEntity(@Context SecurityContext sc, @PathParam(PARAM_ENTITY)
...
-                return new SetDefaultVersionCommand(null, entity, null).execute();
+                return;
             }
         }
-        throw new SecurityException("Unauthorized request for 'DELETE /?entity/default'. One of the following roles is needed: "+roles);
+
+        throw new RoleException(roles);
+    }
+
+    private static class RoleException extends Exception{
```
jewzaam: Why is this a private inner class?  It ends up being thrown as the cause in `SecurityException`.  Either:
1. make this a proper public class or
2. do not use this class as the cause of any exception thrown by public api's or
3. update `checkPermission` to return a boolean to indicate success or failure and not use exception at all

I recommend the 3rd option.

jewzaam:
Moving back to in progress until inner exception class is resolved.

metadata/src/main/java/com/redhat/lightblue/rest/metadata/AbstractMetadataResource.java
dcrissman: The 3rd option was actually my initial implementation, but the problem there is that the calling method would not have access to the required roles for the SecurityException message (to maintain existing functionality). So instead I opted to have checkPermission simply throw the new RoleException. I am happy to make RoleException a proper public class, I didn't initially because of the specific use case.

I think ideally, the core framework would own the checkPermission logic (or some generic variation thereof) so that should we ever add a new "view" layer it would not need to re-implement this. As it is, the SecurityContext needs to be consulted and that seems to be a lightblue-rest element, so solving that seemed to be outside the scope of this issue. In this scenario, I think a RoleException probably would make more sense.

dcrissman:
@jewzaam Changes made that we discussed. Ready for review.

PR #66, opened 2015-01-13
getHttpStatus uses the wrong error list

bserdar:


derek63:
Builds locally, (but not in Travis), merging now

PR #70, opened 2015-01-16
Fixes #41: updated artifactId's

### LINKED ISSUE #41: Rename all artifacts to be prefixed with lightblue-rest
Like was done in lightblue-client https://github.com/lightblue-platform/lightblue-client/pull/34

jewzaam:
requires:
* https://github.com/lightblue-platform/lightblue-core/pull/290
* https://github.com/lightblue-platform/lightblue-audit-hook/pull/23
* https://github.com/lightblue-platform/lightblue-mongo/pull/59
* https://github.com/lightblue-platform/lightblue-rdbms/pull/91


jewzaam:
This fails locally for me as well:
```
rm -rf ~/.m2/repository/com/redhat/lightblue/
mvn dependency:resolve dependency:resolve-plugins -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```

So I'll debug later..

.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
     - chmod 777 ~/onFailure.sh
     - chmod 777 ~/onSuccess.sh
 install:
-    - mvn dependency:resolve dependency:resolve-plugins -DskipTests=true -Dmaven.javadoc.skip=true -B -V
+    - mvn clean install dependency:resolve dependency:resolve-plugins -DskipTests=true -Dmaven.javadoc.skip=true -B -V
```
luan-cestari: I think we could remove the install parameter as in this phase it would just  "... install any dependencies required ..." [1]  and then in the next phase from travis' lifecycle it would "... run a script that effectively runs the build or the tests ..."

[1] http://docs.travis-ci.com/user/build-lifecycle/

.travis.yml
```diff
@@ -11,7 +11,7 @@ before_install:
...
 script:
     - mvn -B clean verify cobertura:cobertura coveralls:cobertura -Dorg.apache.maven.user-settings=~/settings-arquillian.xml  -Dorg.apache.maven.global-settings=~/settings-arquillian.xml
```
luan-cestari: We should change this to install http://docs.codehaus.org/display/MAVENUSER/introduction-to-the-lifecycle

luan-cestari:
@jewzaam Added some comments, I would appreciate your feedback about them

.travis.yml
jewzaam: The issue is that without the install goal dependencies are not available to other modules.  I tried a local build as noted in earlier comments and it fails because crud cannot find common.  If there's some other goal that could be used other than install that's fine, but given the step in the travis config is 'install' this seemed the appropriate place.

jewzaam: http://docs.travis-ci.com/user/build-configuration/#Build-Lifecycle

In this `install` is documented as preparing dependencies.  Adding `-N` would not recurse, but then what's the point?  A fact here is that builds will fail with the way this is setup right now.  Is it really a necessary step?

luan-cestari: We can leave as it is right now, I just wonder if we would have any issue as the install phase will eventually puts the jars produced in the local maven repository and we got an issue during the execution of tests of the script phase (which could be a problem if Travis reuse that host to build for another lightblue's project). Do you also think that problem could happen too?

jewzaam: I suspect that is why this build worked for some PR's and not for others, that a host was reused to build artifacts and therefore were successful.  The artifacts have to be available.  If we want to get away from that we need to manually clean the `.m2` dir probably.  Or not do this step in the install phase and simply rely on the `install` goal to do all the dependency management.

luan-cestari: Agree with you. So, maybe we could change from install to verify goal, so maven would not put it into the repsoitory. The install goal we can use in the line 16 , as it will include the tests. Or how about we use both 'install' goal (from maven) on the install and script phase (from travis) and on failure, we wiped the version installed and move back a previous jar in the local repository  ?

jewzaam:
Pushed an update that removes the `install` section.

luan-cestari:
Seems fine. Just a reminder: the https://github.com/lightblue-platform/lightblue-rest/blob/master/etc/onSuccess.sh will deploy the project (so it will also install locally). That is helpful for other projects so it can install the lightblue side project jars locally , specially the snapshot version)

Issue #27, opened 2014-11-10
Fix travis-ci shrinkwrap issue


jewzaam:
Pull requests cannot build successfully on travis-ci.org and the impact is that we have to merge PR's before we know 100% that it will be a clean build.  This is annoying to say the least.  So far this hasn't resulted in a build failure for master, but it's just a matter of time.

See this for an example: https://travis-ci.org/lightblue-platform/lightblue-rest/jobs/39863929

The result:
```
Results :
Tests in error: 
  com.redhat.lightblue.rest.crud.ITCaseCrudResourceTest: Could not invoke deployment method: public static org.jboss.shrinkwrap.api.spec.WebArchive com.redhat.lightblue.rest.crud.ITCaseCrudResourceTest.createDeployment()
Tests run: 6, Failures: 0, Errors: 1, Skipped: 0
[INFO] ------------------------------------------------------------------------
[INFO] Reactor Summary:
[INFO] 
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-pom . SUCCESS [  2.477 s]
[INFO] lightblue-rest common: com.redhat.lightblue.rest|rest-common SUCCESS [  1.387 s]
[INFO] lightblue-rest: com.redhat.lightblue.rest|auth ..... SUCCESS [  0.235 s]
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-crud  FAILURE [ 12.366 s]
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-crud-basic-auth SKIPPED
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-crud-cert-auth SKIPPED
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-metadata SKIPPED
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-metadata-basic-auth SKIPPED
[INFO] lightblue-rest: com.redhat.lightblue.rest|rest-metadata-cert-auth SKIPPED
[INFO] ------------------------------------------------------------------------
[INFO] BUILD FAILURE
[INFO] ------------------------------------------------------------------------
[INFO] Total time: 17.810 s
[INFO] Finished at: 2014-11-03T18:53:18+00:00
[INFO] Final Memory: 20M/187M
[INFO] ------------------------------------------------------------------------
```

jewzaam:
Fixed with #73

PR #73, opened 2015-01-16
Issue 27 travis arquillian

luan-cestari:
Changing travis.yml
Testing to use some more arguments on both maven and arquillian for a better description of the problem/error. The Arquillian will now log FINEST and maven should say some more information during the Could not invoke deployment method: public static org.jboss.shrinkwrap.api.spec.WebArchive error


luan-cestari:
issue  #27

PR #76, opened 2015-01-19
Cleaned up root dir, moved scripts and testing stuff into etc dir

jewzaam:
And deleted .travis.yml.debug as I can't find any docs about how it's used by travis.  If it's needed I'll add it back in.


luan-cestari:
Cool, how about to replicate the .travis.yml and other files for other repositories? Do you think we could open issues for all other java projects? Or is it better first discuss on http://dev.forum.lightblue.io/ ?

jewzaam:
@luan-cestari if you want to do it I say just do it, unless you don't have time.  In that case I'd just open one issue (takes less time) in core.

PR #80, opened 2015-01-20
Fixes #77: added simple filtered jsp to report the rest app version

### LINKED ISSUE #77: Expose version as rest service
Add /version API to expose the current version.  Full URL's would be:

```
http://hostname:port/rest/metadata/version
http://hostname:port/rest/data/version
```

jewzaam:


jewzaam:
Running failed oraclejdk8 build again.

luan-cestari:
Maybe there could be an integration test for each project to test that new service, what do you think @jewzaam ?

jewzaam:
Add an issue if we want it.  Thought of it but there would be much groundwork to cover for it to be automated.  The arquillian tests we have inject a rest resource, we don't have anything in place I think to do the jsp testing.  It would be nice to add though later.

PR #83, opened 2015-01-23
Wildcard null pointer exception#96

luan-cestari:
Test the fix for https://github.com/lightblue-platform/lightblue-rdbms/issues/96


luan-cestari:
Depends on https://github.com/lightblue-platform/lightblue-rdbms/pull/100

PR #86, opened 2015-02-01
Add support for Cross Origin Resource Sharing (fixes #75)

### LINKED ISSUE #75: Enable cross-origin resource sharing (CORS)
It doesn't appear there is a straightforward way to do this with JBoss, but it is pretty easy to make a generic servlet filter that checks for and adds the appropriate headers. I'll try opening a PR for this.

alechenninger:
If crud or metadata find a lightblue-[crud|metadata]-cors.json on the classpath, they will enable CORS, using configuration derived from the json config, falling back to default values if not all fields are defined explicitly.

There is no configuration included by default, so therefore CORS is _not_ enabled unless a configuration file is explicitly added to the classpath (presumably via puppet and a jboss module like other configs).

An example nonsense json config: 
```json
{
  "urlPatterns": ["/test/*", "/data/*"],
  "allowedOrigins": ["http://test.com", "http://test2.com"],
  "allowedMethods": ["GET", "POST"],
  "allowedHeaders": [
    "Origin",
    "Accept",
    "X-Requested-With"
  ],
  "exposedHeaders": ["test", "exposed", "headers"],
  "preflightMaxAge": 10,
  "allowCredentials": false,
  "enableLogging": true
}
```


pom.xml
```diff
@@ -146,13 +147,48 @@ along with this program.  If not, see <http://www.gnu.org/licenses />.
...
+                <version>1.10.19</version>
+                <scope>test</scope>
+            </dependency>
         </dependencies>
     </dependencyManagement>
     <dependencies>
         <dependency>
             <groupId>junit</groupId>
             <artifactId>junit</artifactId>
-            <scope>test</scope>
```
jewzaam: Why is this scope=test removed from junit?

pom.xml
```diff
@@ -146,13 +147,48 @@ along with this program.  If not, see <http://www.gnu.org/licenses />.
...
+                        <groupId>javax.servlet</groupId>
+                        <artifactId>servlet-api</artifactId>
+                    </exclusion>
+                </exclusions>
+            </dependency>
+            <dependency>
+                <groupId>org.hamcrest</groupId>
+                <artifactId>hamcrest-core</artifactId>
+                <version>1.3</version>
+                <!-- Must leave compile scope to prevent Mockito's transitive, conflicting, runtime
```
jewzaam: Could the conflicting version be excluded here and explicit scope be set instead of a comment nobody will read?

jewzaam:
At a glance looks reasonable.  I wonder at the builder, seems overkill but fine.  I don't know enough about list/array manipulations to speak to if it's done well.  I haven't had time to look at the tests to see what is being covered.

If anybody else wants to pick up on this one for review go for it, don't wait on me.

pom.xml
alechenninger: I just moved it to the dependency management section.

pom.xml
alechenninger: Good call

common/src/main/java/com/redhat/lightblue/rest/cors/CorsConfiguration.java
```diff
@@ -0,0 +1,518 @@
...
+         *
+         * <p>Defaults to "/*".
+         */
+        public Builder urlPatterns(String... patterns) {
+            Objects.requireNonNull(patterns, "patterns");
+
+            String[] copy = new String[patterns.length];
+            System.arraycopy(patterns, 0, copy, 0, patterns.length);
+
+            this.urlPatterns = unmodifiableList(Arrays.asList(copy));
```
dcrissman: Is arraycopy needed if you are putting it in an unmodifiableList?

common/src/main/java/com/redhat/lightblue/rest/cors/EbayCorsFilterRegistration.java
```diff
@@ -0,0 +1,63 @@
+package com.redhat.lightblue.rest.cors;
+
+import org.ebaysf.web.cors.CORSFilter;
+
+import javax.servlet.FilterRegistration;
+import javax.servlet.ServletContext;
+import java.util.Iterator;
+import java.util.List;
+
+public class EbayCorsFilterRegistration implements CorsFilterRegistration {
```
dcrissman: Why Ebay?

common/src/main/java/com/redhat/lightblue/rest/cors/CorsConfiguration.java
alechenninger: The backing list (and its backing array) can still change, unmodifiableList is just a view onto a list that prevents further modifications. An unmodifiable list is unfortunately not immutable so this is why I'm kind of going through hoops here to be clean about it.

common/src/main/java/com/redhat/lightblue/rest/cors/EbayCorsFilterRegistration.java
alechenninger: They have a simple CORS filter already implemented. If we ever adopt KeyCloak we can use that instead.

https://github.com/ebay/cors-filter

common/src/test/java/com/redhat/lightblue/rest/cors/CorsConfigurationBuilderTest.java
```diff
@@ -0,0 +1,96 @@
...
+        assertThat(config.getAllowedHeaders(), contains("Origin", "Accept", "X-Requested-With"));
+        assertThat(config.getAllowedMethods(), contains("GET", "POST"));
+        assertThat(config.getExposedHeaders(), contains("test", "exposed", "headers"));
+        assertThat(config.getPreflightMaxAge(), is(10));
+        assertThat(config.areCredentialsAllowed(), is(false));
+        assertThat(config.isLoggingEnabled(), is(true));
+    }
+
+    @Test
+    public void shoulNotOverrideDefaultsIfNotProvidedInJsonConfiguration() throws IOException {
```
dcrissman: minor typo... missing the 'd'

PR #87, opened 2015-02-03
Add resource filtering to overlay projects

derek63:
So that version is returned from version servlet.  It seems these don't propagate down when pulled in via overlay


jewzaam:
There's got to be something with versions of maven or plugins going on here cause I did check this worked locally before putting the PR in.  Oh well!

PR #95, opened 2015-02-17
Updated travis config to move coveralls to after_script section.  Should...

jewzaam:
... always run but never cause build to fail.

Conflicts:
	.travis.yml


dcrissman:
Build failed because master is broken. This is a low risk change that works in other modules. I will approve so that accurate errors can be provided.

PR #99, opened 2015-02-23
Fixes #17 - Update login module (auth) to cache authz

### LINKED ISSUE #17: Update login module (auth) to cache authz
- [ ] Update to cache a small set of authorizations from LDAP
- [ ] Update hystrix command to use cache if request to LDAP fails
- [ ] Update login module to use cache in some cases (assumes we want to not hit LDAP for every single request but can trust cached data for a period of time)

luan-cestari:
Changes to solve the issue #17

### DIFF
--- a/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
+++ b/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
@@ -15,7 +15,7 @@ import java.util.concurrent.TimeUnit;
  */
 public class LDAPCache {
     private static final Logger LOGGER = Logger.getLogger(LDAPCache.class);
-    private static final LoadingCache<LDAPCacheKey, SearchResult> ldapCacheSession; // non-persisted cache
+    private static final Cache<LDAPCacheKey, SearchResult> ldapCacheSession; // non-persisted cache
     private static final Cache<LDAPCacheKey, List<String>> userRolesCacheSession; // non-persisted cache
 
     // The values should be loaded from somewhere and also we should define the default values/behavior as well
@@ -34,13 +34,7 @@ public class LDAPCache {
                                 LOGGER.debug("This data from " + notification.getKey() + " evacuated due:" + notification.getCause());
                             }
                         }
-                ).build(
-                        new CacheLoader<LDAPCacheKey, SearchResult>() {
-                            public SearchResult load(LDAPCacheKey key) throws NamingException, LDAPUserNotFoundException, LDAPMutipleUserFoundException {
-                                return LDAPSearcher.searchLDAPServer(key);
-                            }
-                        }
-                );
+                ).build();
 
         userRolesCacheSession = CacheBuilder.newBuilder()
                 .concurrencyLevel(10) // handle 10 concurrent request without a problem
@@ -59,7 +53,7 @@ public class LDAPCache {
                 ).build();
     }
 
-    public static LoadingCache<LDAPCacheKey, SearchResult> getLDAPCacheSession() {
+    public static Cache<LDAPCacheKey, SearchResult> getLDAPCacheSession() {
         return ldapCacheSession;
     }
 
--- a/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
+++ b/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
@@ -1,15 +1,19 @@
 package com.redhat.lightblue.rest.auth.ldap;
 
-import com.google.common.cache.*;
+import com.google.common.cache.Cache;
+import com.google.common.cache.CacheBuilder;
+import com.google.common.cache.RemovalListener;
+import com.google.common.cache.RemovalNotification;
 import org.jboss.logging.Logger;
 
+import javax.annotation.ParametersAreNonnullByDefault;
 import javax.naming.directory.SearchResult;
 import java.util.List;
 import java.util.concurrent.TimeUnit;
 
 /**
  * This class hold the references to all local caches and have helper method to deal with them
- *
+ * <p/>
  * Created by lcestari on 2/23/15.
  */
 public class LDAPCache {
@@ -29,7 +33,7 @@ public class LDAPCache {
                                 LOGGER.info("Removal Listener created");
                             }
 
-                            public void onRemoval(RemovalNotification notification) {
+                            public void onRemoval(@ParametersAreNonnullByDefault RemovalNotification notification) {
                                 LOGGER.debug("This data from " + notification.getKey() + " evacuated due:" + notification.getCause());
                             }
                         }
@@ -45,7 +49,7 @@ public class LDAPCache {
                                 LOGGER.info("Removal Listener created");
                             }
 
-                            public void onRemoval(RemovalNotification notification) {
+                            public void onRemoval(@ParametersAreNonnullByDefault RemovalNotification notification) {
                                 LOGGER.debug("This data from " + notification.getKey() + " evacuated due:" + notification.getCause());
                             }
                         }
@@ -60,7 +64,7 @@ public class LDAPCache {
         return userRolesCacheSession;
     }
 
-    public static void invalidateKey(LDAPCacheKey key){
+    public static void invalidateKey(LDAPCacheKey key) {
         ldapCacheSession.invalidate(key);
         userRolesCacheSession.invalidate(key);
     }
--- a/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
+++ b/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
@@ -1,0 +1,74 @@
+package com.redhat.lightblue.rest.auth.ldap;
+
+import com.google.common.cache.*;
+import org.jboss.logging.Logger;
+
+import javax.naming.NamingException;
+import javax.naming.directory.SearchResult;
+import java.util.List;
+import java.util.concurrent.TimeUnit;
+
+/**
+ * This class hold the references to all local caches and have helper method to deal with them
+ *
+ * Created by lcestari on 2/23/15.
+ */
+public class LDAPCache {
+    private static final Logger LOGGER = Logger.getLogger(LDAPCache.class);
+    private static final LoadingCache<LDAPCacheKey, SearchResult> ldapCacheSession; // non-persisted cache
+    private static final Cache<LDAPCacheKey, List<String>> userRolesCacheSession; // non-persisted cache
+
+    // The values should be loaded from somewhere and also we should define the default values/behavior as well
+    static {
+        ldapCacheSession = CacheBuilder.newBuilder()
+                .concurrencyLevel(10) // handle 10 concurrent request without a problem
+                .maximumSize(500) // Hold 500 sessions before remove them
+                .expireAfterWrite(7, TimeUnit.DAYS) // If the session is inactive for more than 7 days, remove it
+                .removalListener(
+                        new RemovalListener<LDAPCacheKey, SearchResult>() {
+                            {
+                                LOGGER.info("Removal Listener created");
+                            }
+
+                            public void onRemoval(RemovalNotification notification) {
+                                LOGGER.debug("This data from " + notification.getKey() + " evacuated due:" + notification.getCause());
+                            }
+                        }
+                ).build(
+                        new CacheLoader<LDAPCacheKey, SearchResult>() {
+                            public SearchResult load(LDAPCacheKey key) throws NamingException, LDAPUserNotFoundException, LDAPMutipleUserFoundException {
+                                return LDAPSearcher.searchLDAPServer(key);
+                            }
+                        }
+                );
+
+        userRolesCacheSession = CacheBuilder.newBuilder()
+                .concurrencyLevel(10) // handle 10 concurrent request without a problem
+                .maximumSize(500) // Hold 500 sessions before remove them
+                .expireAfterWrite(7, TimeUnit.DAYS) // If the session is inactive for more than 7 days, remove it
+                .removalListener(
+                        new RemovalListener<LDAPCacheKey, List<String>>() {
+                            {
+                                LOGGER.info("Removal Listener created");
+                            }
+
+                            public void onRemoval(RemovalNotification notification) {
+                                LOGGER.debug("This data from " + notification.getKey() + " evacuated due:" + notification.getCause());
+                            }
+                        }
+                ).build();
+    }
+
+    public static LoadingCache<LDAPCacheKey, SearchResult> getLDAPCacheSession() {
+        return ldapCacheSession;
+    }
+
+    public static Cache<LDAPCacheKey, List<String>> getUserRolesCacheSession() {
+        return userRolesCacheSession;
+    }
+
+    public static void invalidateKey(LDAPCacheKey key){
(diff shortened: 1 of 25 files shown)


luan-cestari:
The PR https://github.com/lightblue-platform/lightblue-core/pull/335 might fix the problem on Travis using openjdk7

luan-cestari:
This PR can't be merged due the RDBMS module is outdated. I already made a PR to fix a part of the problem (the Hystrix thing which impact not only rdbms but core and mongo repositories as well)  (https://github.com/lightblue-platform/lightblue-rdbms/pull/111) , but rdbms is quite outdated so there are still things to change to make compatible with the lightblue new version.

luan-cestari:
Pending PR https://github.com/lightblue-platform/lightblue-mongo/pull/78 to be merged to rebuild this PR

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPMutipleUserFoundException.java
```diff
@@ -0,0 +1,11 @@
+package com.redhat.lightblue.rest.auth.ldap;
+
+/**
+ * This simple exception is used when an unexpected behavior is found from the LDAP server where more than one entry is returned
+ *
+ * Created by lcestari on 2/23/15.
+ */
+public class LDAPMutipleUserFoundException extends Exception {
+    public LDAPMutipleUserFoundException(String message) {
```
jewzaam: Loose the message here, why?

luan-cestari: My bad. I asked the IDEA to generate the Exception and I didn't release it didn't used the super constructor. Fixing

luan-cestari: Done

luan-cestari:
Note: last build worked fine. I will rebuild it to see if happens the issue https://github.com/lightblue-platform/lightblue-rest/issues/103

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
```diff
@@ -1,62 +1,122 @@
...
+        } else if(Strings.isNullOrEmpty(ldapSearchBase)) {
+            throw new IllegalArgumentException(String.format(INVALID_PARAM, "ldapSearchBase"));
+        } else if(ldapContext == null) {
+            throw new IllegalArgumentException(String.format(INVALID_PARAM, "ldapContext"));
+        }
 
-        this.ldapContext = ldapContext;
-        this.ldapSearchBase = ldapSearchBase;
-        this.uid = uid;
+        this.cacheKey = new LDAPCacheKey(uid, ldapContext, ldapSearchBase,  "(uid=" + uid + ")");
```
jewzaam: Logic for creating a new cache key is duplicated.  Maybe a static method on LDAPCacheKey that takes uid, ldapContext, and ldapSearchBase as arguments and creates a new instance would be good.  Make the constructor protected.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCacheKey.java
```diff
@@ -0,0 +1,58 @@
...
+    final String ldapSearchBase;
+    final String searchFilter;
+    final SearchControls searchControls = new SearchControls();
+
+    public LDAPCacheKey(String uid, LdapContext ldapContext, String ldapSearchBase, String searchFilter) {
+        this.uid = uid;
+        this.ldapContext = ldapContext;
+        this.ldapSearchBase = ldapSearchBase;
+        this.searchFilter = searchFilter;
+        this.searchControls.setSearchScope(SearchControls.SUBTREE_SCOPE);
```
jewzaam: Provide a constructor that takes search scope as an argument so it could be changed?  In our deployment this makes sense, will it make sense everywhere?

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
```diff
@@ -0,0 +1,74 @@
...
+    private static final Logger LOGGER = Logger.getLogger(LDAPCache.class);
+    private static final LoadingCache<LDAPCacheKey, SearchResult> ldapCacheSession; // non-persisted cache
+    private static final Cache<LDAPCacheKey, List<String>> userRolesCacheSession; // non-persisted cache
+
+    // The values should be loaded from somewhere and also we should define the default values/behavior as well
+    static {
+        ldapCacheSession = CacheBuilder.newBuilder()
+                .concurrencyLevel(10) // handle 10 concurrent request without a problem
+                .maximumSize(500) // Hold 500 sessions before remove them
+                .expireAfterWrite(7, TimeUnit.DAYS) // If the session is inactive for more than 7 days, remove it
```
jewzaam: Need an issue to track making the cache configuration not be hard coded.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
```diff
@@ -1,62 +1,122 @@
...
+
+        public FallbackViaLDAPServerProblemCommand(LDAPCacheKey cacheKey, Throwable failedExecutionException) {
+            super(HystrixCommand.Setter.withGroupKey(HystrixCommandGroupKey.Factory.asKey(GROUPKEY)).
+                    andCommandKey(HystrixCommandKey.Factory.asKey(GROUPKEY + ":" + FallbackViaLDAPServerProblemCommand.class.getSimpleName())));
+            this.cacheKey = cacheKey;
+            this.failedExecutionException = failedExecutionException;
+        }
+
+        @Override
+        protected SearchResult run() throws Exception {
```
jewzaam: 1. why is this attempting to reach back out to ldap?
2. why isn't this always returning whatever is cached?

I think a retry of the request might be good, but not sure it should be in the fallback.  There's some level of duplication of logic with the public command though reuse is a problem.  My preference would be to keep it simple.  Meaning, the fallback for `LdapFindUserByUidCommand` simply returns whatever is cached.  If nothing is cached it fails in some way.  Either throw the original exception, a new exception, or return null.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
```diff
@@ -1,62 +1,122 @@
...
+
+            } else if(failedExecutionException instanceof Exception) {
+                // if it is an exception, propagate it until someone handle it properly
+                throw ((Exception)failedExecutionException);
+            } else {
+                // probably java.lang.Error or a custom Throwable exception. In this case we can't handle the problem
+                // and if we wrap it we can change the JVM to inconsistent state, so we have to exit with error status
+                // code
+                LOGGER.fatal("a Throwable instance (which is not instance of Exception) was thrown and it can't be handled.",failedExecutionException);
+                System.exit(-1);
```
jewzaam: Why is there a system exit here?  That will exit the JVM.

luan-cestari: Yeah, that was the most uncomfortable but necessary thing, I tried to briefly explain why I did that in the comment, and again in summary as we are dealing with Throwable and we tried all the Exceptions classes and related possibility, that can be an Error which is something wrong with the JVM or another not recoverable issue. Anotehr problem with this that lead em to do that is that Hystrix throws in the fallback back Throwable and in the run method it only accepts Execptions, making us to deal with the Throwable (as we can not re throw it by java lang spec) and if we wrap with an Execption, again, we can be dealing with out of memory and ignoring it would lead the JVM to not predictable state , e.g.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCache.java
luan-cestari: done https://github.com/lightblue-platform/lightblue-rest/issues/109

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LDAPCacheKey.java
luan-cestari: done

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
luan-cestari: OK. I will remove the retry from cache ( I sort of thought that would be nice to retry but might need a more careful plan how we would do that)

luan-cestari: Done

luan-cestari:
Made some enhanced around the exceptions and created TODO REVIEWER  for points we can discuss more here

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
jewzaam: https://github.com/lightblue-platform/lightblue-rest/issues/112 created for this

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
```diff
@@ -1,62 +1,132 @@
...
+
+        public FallbackViaLDAPServerProblemCommand(LDAPCacheKey cacheKey, Throwable failedExecutionThrowable) {
+            super(HystrixCommand.Setter.withGroupKey(HystrixCommandGroupKey.Factory.asKey(GROUPKEY)).
+                    andCommandKey(HystrixCommandKey.Factory.asKey(GROUPKEY + ":" + FallbackViaLDAPServerProblemCommand.class.getSimpleName())));
+            this.cacheKey = cacheKey;
+            this.failedExecutionThrowable = failedExecutionThrowable;
+        }
+
+        @Override
+        protected SearchResult run() throws Exception {
```
jewzaam: Why doesn't FallbackViaLDAPServerProblemCommand always fall back on the cache?  Why would we cause all requests to fail if we have some valid cached roles for the principal?  I think in the fallback we should only propagate up errors when we don't have something to return from the cache.

crud/src/test/resources/ITCaseCrudResourceRDBMSTest/config/config.properties
```diff
@@ -1,3 +1,4 @@
 # set default isolation strategy for tests to SEMAPHORE else breakpoints in IDE
 # when debugging unit tests will timeout with default strategy (THREAD).
 hystrix.command.default.execution.isolation.strategy=SEMAPHORE
+execution.isolation.thread.timeoutInMilliseconds=30000
```
jewzaam: `execution.isolation.thread.timeoutInMilliseconds` isn't a valid property I think.  Anyway, the default isolation is SEMAPHORE in tests, not THREAD, so this wouldn't have any impact.  See https://github.com/Netflix/Hystrix/wiki/Configuration#execution.isolation.thread.timeoutInMilliseconds

pom.xml
```diff
@@ -257,6 +250,12 @@ along with this program.  If not, see <http://www.gnu.org/licenses />.
         <sonar.host.url>http://127.0.0.1:8080</sonar.host.url>
         <sonar.projectName>lightblue-rest</sonar.projectName>
         <rpm.install.basedir>/usr/share/jbossas/standalone/deployments</rpm.install.basedir>
+        <guava>18.0</guava>
```
jewzaam: Because of how the release plugin works it's preferred to put the version in the dependency directly instead of in a property.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
luan-cestari: It relies on the cache when It couldn't connect to the LDAP server (which I think it is the use case). I didn't understand why would all the request would fail as you mentioned, unless you are talking about Error from the JVM. About the Exception propagation,  roughly it is doing that , the other propagations that you can see in that method are for cases that we don't handle but Hystrix & Java best practices suggest to us to propagate or even exist the process

crud/src/test/resources/ITCaseCrudResourceRDBMSTest/config/config.properties
luan-cestari: Yeah, I agree with you but I saw a timeout in one of the tests, which I could suppose that the Hystrix couldn't find this file in the classpath or it ignored the first line. I put this to check the second hypothesis, it didn't happen again with me or in travis, so I don't know if it worked / fixed . I can remove it but I would suggest to we have an issue to create a test and force an extreme scenario to see if have any issue (would have an special profile to run only occasionally). What do you think?

pom.xml
luan-cestari: Done

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
jewzaam: It seems like over-engineering.  I want to see the fallback just fall back to the cache.  We don't do any special handling for other cases in code anywhere else.  I especially am worried about the System.exit(-1).  If the JVM gets into such a broken state we could expect other code that doesn't use a fallback to have issues that would propagate to be handled.  Additionally, if there are such conditions that shouldn't go to a fallback I would think HystrixCommand should handle it. If it doesn't we could put an issue in for them to fix it.  We should expect that if fallback is called that we can do the fallback behavior we need, in this case try to get data from the cache.

@bserdar thoughts?

crud/src/test/resources/ITCaseCrudResourceRDBMSTest/config/config.properties
jewzaam: The property `execution.isolation.thread.timeoutInMilliseconds` is not a valid configuration for hystrix based on the doc.  Per documentation the property name to configure the default behavior is `hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds`.

luan-cestari: Removed that KV

luan-cestari:
Merged with the remote master

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
luan-cestari: I opened an [issue on hystrix](https://github.com/Netflix/Hystrix/issues/713) about this asking for help (and if it is a bug). I all the wiki pages and some Hystrixs' classes and I think there is another way to handle it (maybe with observable or plugin) but it really seems that the default behavior Hystrix would wrap the Errors (or anything that isn't an Exception) (which can be not desired behavior in JVM, I put some references even from Java lang spec in the issue) .

luan-cestari:
Created issue https://github.com/lightblue-platform/lightblue-rest/issues/113 so we can track the Hystrix problem (more information on the issue itself)

PR #102, opened 2015-02-24
Fixes #101 - Change Arquillian to use embedded wildfly

### LINKED ISSUE #101: Change Arquillian  to use embedded wildfly
Now it is using the most basic Arquillian container that helps with CDI but it might be interresting to change to use wildfly which is much more close to the application server that we are going to use to deploy. I thought this working on https://github.com/lightblue-platform/lightblue-rest/issues/2 where I needed to test the client side of the rest integration but the current arquillian container doesn't provide this feature. I think this issue might also help the old issue/discussion to make lightblue to work on  wildfly server (  http://forum.lightblue.io/Unable-to-deploy-lightblue-on-wildfly-td4.html#a7 )

Container Injection Support Matrix comparison:
https://docs.jboss.org/author/display/ARQ/WildFly+8.1.0+-+Embedded
https://docs.jboss.org/author/display/ARQ/Weld+EE+1.1+-+Embedded

https://docs.jboss.org/author/display/ARQ/Container+adapters
https://github.com/aslakknutsen/arquillian-showcase

luan-cestari:
This is a draft of the solution., it doesn't aim to be fully working as this issue is new and it needs to be reviewed and set the right priority. See more information about the issue on  https://github.com/lightblue-platform/lightblue-rest/issues/101

### DIFF
--- a/metadata/pom.xml
+++ b/metadata/pom.xml
@@ -92,6 +92,48 @@
                     </webResources>
                 </configuration>
             </plugin>
+            <plugin>
+                <groupId>org.apache.maven.plugins</groupId>
+                <artifactId>maven-dependency-plugin</artifactId>
+                <version>2.8</version>
+                <executions>
+                    <execution>
+                        <id>unpack</id>
+                        <phase>process-test-classes</phase>
+                        <goals>
+                            <goal>unpack</goal>
+                        </goals>
+                        <configuration>
+                            <artifactItems>
+                                <artifactItem>
+                                    <groupId>org.wildfly</groupId>
+                                    <artifactId>wildfly-dist</artifactId>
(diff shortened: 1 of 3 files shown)


alechenninger:
I just got this working (sort of). Few things I noticed you might need:

- web.xml (since this is a real servlet/EE container not just CDI)
- When ProxyFactory is used, I think this will probably fail... looks like must use with interface (and all we have is class or abstract class)

alechenninger:
Also, you can use URLConnectionClientExecutor instead of apache and avoid some complexity I think

davidlanouette:
Since this PR is over 7 years old, I'm going to close it.  If you want to resubmit the change, please rebase on master, address any issues, and create a new PR.

PR #110, opened 2015-03-09
Fix coveralls

luan-cestari:
Fix the problem about coveralls that runs on travis during after_script phase. More info https://github.com/lightblue-platform/lightblue-rest/issues/104


jewzaam:
@luan-cestari build fails

luan-cestari:
@jewzaam thanks. This PR is depending some changes from https://github.com/lightblue-platform/lightblue-rest/pull/99 (like the other PR in the rest repository that also failed due the same reason) . I think we can move back to ready_for_review and someone could review and merge that other PR.

PR #114, opened 2015-03-13
Create classes jars for crud.war and metadata.war

paterczm:
This allows bootstrapping lightblue rest layer without a need of a container which understands web archives.


luan-cestari:
Build failed but due other issues. I will make travis build it again but as the plugin's web site mention (https://maven.apache.org/plugins/maven-war-plugin/faq.html) this would not impact us, it will just deploy both a war and a jar file

Issue #115, opened 2015-03-13
Fix Hystrix configuration


luan-cestari:
Found strange behavior case when the commands were taking more than one second to respond and after looking into the Hystrix documentation (which  was outdated https://github.com/Netflix/Hystrix/issues/714 ) and code I understood that Hystrix changed its default behavior. After search more a bit found the https://github.com/Netflix/Hystrix/blob/master/CHANGELOG.md that make it clear that now even semaphore isolation strategy will timeout. I opened an issue https://github.com/Netflix/Hystrix/issues/714 asking if they could update the documentation and even change that property name and they are working on it. On meanwhile, we need to make sure that every project that relys on Hystrix have its configuration file with https://github.com/Netflix/Hystrix/wiki/Configuration#execution.isolation.thread.timeoutInMilliseconds property with the reasonable amount of time that you would expect that after that threshold it should fire a time out exception from Hystrix. (Note: so this also apply for our puppet scripts that may need to specify the  timeout  parameter as well)  .

jewzaam:
Nice find, I added issues to ldap and mongo, as they use this isolation strategy for tests as well.

luan-cestari:
Thanks. O think they will also rename that property, so It would be ncie that we leave at least one issue one about this to make sure that we have the right property set (as we can get a newer version during lightblue release and have a simialr problem once again)

luan-cestari:
The issue in Hystrix about the renaming -> https://github.com/Netflix/Hystrix/issues/673

Issue #116, opened 2015-03-17
LDAP Authorization cache doesn't appear to be working


derek63:
The changes implemented in https://github.com/lightblue-platform/lightblue-rest/pull/99 were recently deployed to our dev environment, and we lost all ability to access any of the REST endpoints (403 returned from JBoss in all cases).  Upon investigation, no information was being written to the access.log file, so it seemed like LDAP wasn't even being consulted for user roles.  I reverted the login module code to the version prior to the above referenced PR, and was able to access the services again.  So it appears that something related to authorization cache is not working correctly.  I didn't debug in detail, as I was primarily focused on restoring service in dev, so @luan-cestari, can you take a look at this when you have a chance?  Perhaps we can replicate the behavior in a unit test using an LDAP instance in order to get it resolved.

luan-cestari:
@derek63 What about I make a PR with more log statements so we  could have a better picture of the problem and i parallel I will also work creating integration tests to this module (which will take more time to finish)?

derek63:
Sure, more logging statements could help to isolate where the issue lies, and we can use that to troubleshoot.  Let me know when you get that created and I can merge it in.

luan-cestari:
Created the PR for the log. Feel free to review it  & merge it -> https://github.com/lightblue-platform/lightblue-rest/pull/117/files

luan-cestari:
New PR for this problem -> https://github.com/lightblue-platform/lightblue-rest/pull/119

PR #117, opened 2015-03-17
Ldap log#116

luan-cestari:
Related to this issue https://github.com/lightblue-platform/lightblue-rest/issues/116


auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapFindUserByUidCommand.java
```diff
@@ -78,7 +84,7 @@ This methods is executed for all types of failure such as run() failure, timeout
      */
     @Override
     protected SearchResult getFallback() {
-        LOGGER.info("Error during the execution of the command. Falling back to the cache");
+        LOGGER.error("Error during the execution of the command. Falling back to the cache");
```
jewzaam: Should we log this at error level or something lower, maybe warn?  Because we have a fallback, seems it could be a warning and if the fallback fails we'd have an error.

luan-cestari: Agree. I will switch to warn level . At the first sight I thought this would not impact being this way but as it can have a heavy load it could wast a lot of space and cpu .

jewzaam:
Build failing, maybe because of https://github.com/lightblue-platform/lightblue-rest/issues/115

jewzaam:
Put in https://github.com/lightblue-platform/lightblue-rest/pull/118 to hopefully fix #115 and get rid of the timeout issue.

luan-cestari:
rebuilding after the PR been merged

PR #119, opened 2015-03-18
Fixes #116 - Ldap integration test

### LINKED ISSUE #116: LDAP Authorization cache doesn't appear to be working

luan-cestari:
Related to this issue https://github.com/lightblue-platform/lightblue-rest/issues/116 (this PR also incorporate the changes of https://github.com/lightblue-platform/lightblue-rest/pull/117 )


luan-cestari:
Fixed some log  dependency issues. Fixed the possible problem with cache that we found (but we didn't get evidences of the root cause). It need to finsih an unit test but I don't have enough knowledge about LDAP and Apache DS API, if someone could help with that, I can update this PR

luan-cestari:
Some references that might be helpful: https://stackoverflow.com/questions/26398046/adding-a-new-attribute-to-an-object-class-and-expecting-it-to-automatically-show   http://jewzaam.gitbooks.io/lightblue-specifications/content/backend_specifications/ldap/crud_operation_examples.html   https://stackoverflow.com/questions/1560230/running-apache-ds-embedded-in-my-application   https://svn.apache.org/repos/asf/directory/apacheds/tags/1.5.3/core-integ/src/test/java/org/apache/directory/server/core/partition/PartitionIT.java    https://stackoverflow.com/questions/10353061/embedded-apacheds-in-my-application-it-fails-on-service-startup-defaultsch

luan-cestari:
The unit test remaining just need to create a valid object into LDAP server to be queried

jewzaam:
@luan-cestari does this fix #116?  And is there any more work to do before review? Your last comment makes me think something needs done with a unit test still.

luan-cestari:
@jewzaam it probably solve it as we don't have enough data about the issue, so I wouldn't guarantee it solves the problem. The remaining test is something good to have, I ping some people on the channel for help, I think Derek is working on it.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProvider.java
```diff
@@ -54,16 +53,21 @@ public LightblueLdapRoleProvider(String server, String searchBase, String bindDn
...
 
     @Override
     public List<String> getUserRoles(String userName) {
+        LOGGER.debug("Invoking LightblueLdapRoleProvider#getUserRoles");
         List<String> userRoles = new ArrayList<>();
-
         try {
-            userRoles.addAll(getUserRolesFromCache(userName));
+            List<String> userRolesFromCache = getUserRolesFromCache(userName);
+            if( userRolesFromCache != null && userRolesFromCache.isEmpty() ) {
```
jewzaam: Condition should be `!userRolesFromCache.isEmpty()` (missing !)

jewzaam:
I'll merge after condition for using roles from cache is update.
Per note in this PR and @derek63 working on something for the unit test I am skipping review of the test.  I'll expect a separate PR coming with heavy updates.

luan-cestari:
Fixed the line commented by @jewzaam

PR #122, opened 2015-03-19
Refactor LightblueLdapRoleProviderTest to use lightblue-ldap-test and ...

derek63:
...UnboundID in-memory database

### DIFF
--- a/auth/src/test/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProviderTest.java
+++ b/auth/src/test/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProviderTest.java
@@ -84,20 +84,6 @@ public class LightblueLdapRoleProviderTest {
     }
 
     @Test
-    @Ignore
-    public void testInMemoryLdapServer() throws Exception {
-        try {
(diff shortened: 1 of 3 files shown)


auth/src/test/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProviderTest.java
```diff
@@ -1,252 +1,117 @@
...
-
-        List<String> userRoles = provider.getUserRoles("nameHere");
-        assertNotNull(userRoles);
-        assertEquals("getUserRoles method should return an empty list",0, userRoles.size());
+    public LightblueLdapRoleProviderTest() throws Exception {
+        super();
     }
 
     @Test
     @Ignore
```
luan-cestari: Do you really intent that this test must be ignored (I think the @Ignore was forgot there from by mistake) ?

derek63: Actually I meant to take out testInMemoryLdapServer altogether, as it doesn't test anything in LightblueLdapRoleProvider.  Removing now

PR #128, opened 2015-04-04
Fixes #120: Log/audit access for REST requests 

### LINKED ISSUE #120: Log/audit access  for REST requests
We have access logs but it only tracks AuthN/Z.  Once authorized we don't track what the client is requesting.  I'd like to see logs to track the principal along with the rest request at least including:
* timestamp
* principal
* rest resource (crud vs metadata)
* operation (insert, update, find, etc)
* entity name
* entity version

From a security point of view we have to be careful not to log:
* query parameters
* request body

As both of these can contain data that must not be exposed.  Query params would be for simple find and a handful of metadata operations.

Options:
* log to file, which can go to log aggregation (such as Splunk)
* create hook

Problem with the hook is we don't fire hooks on metadata operations, only data (crud) operations.  Additionally, we have log aggregation in place already that allows us to visualize data nicely.

Some example logs:

```
2015-03-18 14:26:42,645 principal=lb-consistency-checker resource=crud operation=find entityName=migrationJob entityVersion=1.2.0
```

Maybe some more info about the request would be helpful, such as:
* request body size
* response body size
* time it took to respond in msec.

luan-cestari:
PR for the issue https://github.com/lightblue-platform/lightblue-rest/issues/120

### DIFF
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -57,5 +57,9 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.github.mifmif</groupId>
+            <artifactId>generex</artifactId>
+        </dependency>
     </dependencies>
 </project>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -22,11 +22,11 @@
     <parent>
         <groupId>com.redhat.lightblue.rest</groupId>
         <artifactId>lightblue-rest-pom</artifactId>
-        <version>1.3.0-SNAPSHOT</version>
+        <version>1.4.0-SNAPSHOT</version>
     </parent>
     <artifactId>lightblue-rest-common</artifactId>
     <packaging>jar</packaging>
-    <version>1.3.0-SNAPSHOT</version>
+    <version>1.4.0-SNAPSHOT</version>
     <name>lightblue-rest: ${project.groupId}|${project.artifactId}</name>
     <dependencies>
         <dependency>
--- a/common/pom.xml
+++ b/common/pom.xml
@@ -24,7 +24,6 @@
         <artifactId>lightblue-rest-pom</artifactId>
         <version>1.3.0-SNAPSHOT</version>
     </parent>
-    <groupId>com.redhat.lightblue.rest</groupId>
     <artifactId>lightblue-rest-common</artifactId>
     <packaging>jar</packaging>
     <version>1.3.0-SNAPSHOT</version>
@@ -53,6 +52,10 @@
         <dependency>
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
+        </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
         </dependency>
     </dependencies>
 </project>
--- a/common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java
+++ b/common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java
@@ -71,9 +71,6 @@ public class LightblueAuditServletFilter implements Filter {
             HttpServletResponse httpServletResponse = (HttpServletResponse) res;
             HttpServletResponse httpServletResponseWrapperBuffered = new HttpServletResponseWrapperBuffered(httpServletResponse,new ByteArrayPrintWriter());
             res = httpServletResponseWrapperBuffered;
-        }
-
-        if(auditReqFlag) {
             stopwatch = Stopwatch.createStarted();
         }
 
--- a/common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java
+++ b/common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java
@@ -27,67 +27,6 @@ import java.util.regex.Pattern;
 @WebFilter(urlPatterns = { "/*" }) // Handle any request
 public class LightblueAuditServletFilter implements Filter {
 
-    //GET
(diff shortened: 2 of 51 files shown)


common/src/main/java/com/redhat/lightblue/rest/audit/LogEntry.java
```diff
@@ -0,0 +1,75 @@
...
+import java.security.Principal;
+
+/**
+ * Created by lcestari on 4/1/15.
+ */
+public class LogEntry {
+    private final String timestampText;
+    private final Principal principal;
+    private final String resource; // /crud or /metadata
+    private final String operation; //operation (insert, update, find, etc)
```
dcrissman: Could this be an enum?

common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java
```diff
@@ -0,0 +1,527 @@
...
+            );
+            LOGGER.info(logEntryString);
+        }
+
+        if(LOGGER.isDebugEnabled()) {
+            LOGGER.debug("LightblueAuditServletFilter.doFilter invoked - end");
+        }
+    }
+
+    protected void setOperationEnittyVersionStatus(HttpServletRequest hReq, boolean isMetadata, LogEntryBuilder logEntryBuilder) {
```
dcrissman: These method seems non-trivial. I wonder if we couldn't break it into smaller methods?

My other thought is maybe we create a single instance of Matcher, before the switch statement?  I am not typically a fan of re-using a variable, but maybe in this case it would help simplify? Then 'else if' statements could be used instead of embedded conditions.

Just thoughts, open for comments.

luan-cestari: Not sure if I got the idea you mentioned about the single instance of Matcher. 

Do you mean create an variable for each case up front (like, before the switch statement) so then I can use 'else if' statement?  In this case, I`m a bit concerned about it could create unnecessary Matcher objects as many cases would not be evaluated during a single request processing (which would give more work to the GC after some requests) , specially as this filter will handle both the data and metadata requests. If GC pause/impact is not a problem, I don`t mind we change to this way.

dcrissman: No, I meant something like this...
```java
Matcher m;
if((m = pattern1.matcher("")).matches()){
  //do something
}
else if((m = pattern2.matcher("")).matches()){
  //do something else
}
//etc

Comment (luan-cestari) [inline common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java]: Good idea. I will update the PR. I will also include the enum that you suggested (but that will not block if someone create another rest service or change an existing one,  due this so tightly coupled nature I even thought to create this class using reflection but I thought it would be overengineering  )

**Comment by luan-cestari** [comment]
Made the changes to use enum for the operation extraction. I thought to use even a loop throw the values of each enum or something but that would probably would make things more complicated as the order of the enums would matter. Ready for review

File: common/pom.xml [hunk 3]
```diff
@@ -54,5 +53,13 @@
             <groupId>org.mockito</groupId>
             <artifactId>mockito-core</artifactId>
         </dependency>
+        <dependency>
+            <groupId>org.slf4j</groupId>
+            <artifactId>slf4j-simple</artifactId>
+        </dependency>
```
dcrissman: Do you need this dependency for testing? It looks like the slf4j-api is already available.

common/pom.xml
```diff
@@ -54,5 +53,13 @@
...
+        <dependency>
+            <groupId>com.github.mifmif</groupId>
+            <artifactId>generex</artifactId>
+        </dependency>
```
dcrissman: test scope also?

common/pom.xml
dcrissman: oh, I see this is set to test scope in the parent pom.

common/pom.xml
dcrissman: oh, I see this is set to test scope in the parent pom.

dcrissman:
A few of the class names start with 'My', is there a better name we could use?

common/src/main/java/com/redhat/lightblue/rest/audit/LightblueAuditServletFilter.java
```diff
@@ -0,0 +1,318 @@
...
+
+
+            HttpServletResponse httpServletResponse = (HttpServletResponse) res;
+            HttpServletResponse httpServletResponseWrapperBuffered = new HttpServletResponseWrapperBuffered(httpServletResponse,new ByteArrayPrintWriter());
+            res = httpServletResponseWrapperBuffered;
+        }
+
+        if(auditReqFlag) {
+            stopwatch = Stopwatch.createStarted();
+        }
```
dcrissman: Can these two if statements be merged?

dcrissman:
I feel like the LogEntryBuilder might be overkill in this case. Does anyone else have any thoughts on this?

luan-cestari:
>A few of the class names start with 'My', is there a better name we could use?
I can rename them. It was anonymous class and when I pulled out that was the suggestion from my IDE. Do you prefer the 'Test' as a prefix?

>I feel like the LogEntryBuilder might be overkill in this case. Does anyone else have any thoughts on this?
It is if you think how complex the LogEntry is, but what I tried to express there is the immutability of the LogEntry, so it would be very consistent. Also the IDE can generate the Builder pattern , so the maintainability is pretty low if we count these facts.

I will merge the two statements and update here. Please, let me know if I forgot to discuss any point you commented

dcrissman:
> I can rename them. It was anonymous class and when I pulled out that was the suggestion from my IDE. Do you prefer the 'Test' as a prefix?

I think we use Fake in other places as not to confused with a suite of tests.

> Regarding LogEntryBuilder

I see what you are trying to do. I am fine with it.

PR #134, opened 2015-04-17
Fixes lightblue-platform/lightblue-core#376 : NPE when performing find on entity with hook

luan-cestari:
Depends on https://github.com/lightblue-platform/lightblue-audit-hook/pull/33

The issue https://github.com/lightblue-platform/lightblue-core/issues/376 have more details

### DIFF
--- a/crud/pom.xml
+++ b/crud/pom.xml
@@ -90,6 +90,11 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.redhat.lightblue.hook</groupId>
+            <artifactId>lightblue-audit-hook</artifactId>
+            <scope>test</scope>
+        </dependency>
     </dependencies>
     <build>
         <plugins>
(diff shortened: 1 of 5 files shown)


crud/pom.xml
```diff
@@ -90,6 +90,11 @@
             <groupId>org.slf4j</groupId>
             <artifactId>slf4j-simple</artifactId>
         </dependency>
+        <dependency>
+            <groupId>com.redhat.lightblue.hook</groupId>
+            <artifactId>lightblue-audit-hook</artifactId>
+            <scope>test</scope>
+        </dependency>
```
dcrissman: Doesn't this create a circular dependency between core and audit-hook?

bserdar: If that's  lightblue-rest/crud/pom.xml, it shouldn't.

dcrissman: @bserdar you're right. I saw lightblue-core in the subject and it threw me. Disregard.

luan-cestari: The build worked fine (as maven wouldn't allow circular dependency as far as I know)

Issue #132, opened 2015-04-07
Refactor LDAP Cache stuff


jewzaam:
We removed it in prep for Ana because it caused enough threads to be consumed that JBoss started having OutOfMemoryError thrown.  I think, based on jstack output, that we have something keeping connections to ldap open.  Therefore, there are two main points to this issue:
1. why would we keep creating and not garbage collecting InitialLdapContext objects?
2. why is the LdapContext saved as part of the LDAPCacheKey?

Actions:
* remove LdapContext from LDAPCacheKey
* only create LdapContext when necessary to lookup something from ldap (meaning a cache miss)
* unit test that the cache is actually working

The last point is key.  We have caused JBoss to tip over using a single principal.  There should have been one lookup to LDAP, cached that lookup, and then used it for subsequent requests.  We should not have found this bug!  But we did, which begs the question why.  Is the cache not really being used?

jewzaam:
Fixed in https://github.com/lightblue-platform/lightblue-rest/pull/135

PR #135, opened 2015-04-20
Rewrite ldap cache to cache roles rather than ldap context

paterczm:
https://github.com/lightblue-platform/lightblue-rest/issues/132

There are 2 caches: rolesCache and fallbackRolesCache. Roles put in the former are evicted after few minutes, so that role membership changes in ldap are reflected reasonably quickly. Roles put in the latter are evicted after few hours. FallbackRolesCache is read only when ldap failure is identified (circuit breaker pattern).


paterczm:
The roles cache expires after 5 minutes. This is the maximum time required for roles changes to take effect. If it's ok for this to take longer, I could increase expiry time and improve performance.

PR #138, opened 2015-04-23
Fix the new audit problem

luan-cestari:
Depends on the PR https://github.com/lightblue-platform/lightblue-audit-hook/pull/34 to work


luan-cestari:
This PR is basically an enhancement for the CRUD test to include the audit hook and check if it works

PR #146, opened 2015-04-30
Fixes lightblue-platform/lightblue-audit-hook#38 : Hook problem - audit currently records all fields during an update, not just the modified fields

luan-cestari:
The issue https://github.com/lightblue-platform/lightblue-audit-hook/issues/38


luan-cestari:
Depends on https://github.com/lightblue-platform/lightblue-mongo/pull/113

dcrissman:
Is this essentially a duplicate of #144?

luan-cestari:
@dcrissman it is essentially the same, but I would think as an enhancement and it also necessary with teh fix on the core&mongo repo

dcrissman:
... but this reverts changes made in #144.

luan-cestari:
@dcrissman , agree partially, I think the #144 removed an verification if the audit was found exactly was we excepted. The changes from that PR that I could apply in this one is about audit metadata such as remove the file and use again " FileUtil.readFile("metadata/audit.json"); " .  Is it OK for you if we do this way?  Do you have other concerns about this PR?

dcrissman:
I removed that because I don't feel it was really testing the audit hook functionality. But I won't block you if you put it back in.

luan-cestari:
Updated the PR

crud/src/test/resources/ITCaseCrudResourceTest/config/config.properties
```diff
@@ -1,4 +1,4 @@
 # set default isolation strategy for tests to SEMAPHORE else breakpoints in IDE
 # when debugging unit tests will timeout with default strategy (THREAD).
 hystrix.command.default.execution.isolation.strategy=SEMAPHORE
-hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds=60000
+hystrix.command.default.execution.isolation.thread.timeoutInMilliseconds=120000
```
dcrissman: Why the increase here?

luan-cestari: Increased to debug many different places without hystrix interrupting the thread.  Can be reverted if you see a big issue (which I would suggest to we make explicit note on the test that we have time constraint). Note that the tests were used to have unlimited time before hystrix 1.4.0

PR #148, opened 2015-05-08
Creating InitialLdapContext on every call to ldap

paterczm:
Because it's likely to be responsible for ```Naming problem with LDAP for user: <user>: javax.naming.CommunicationException: connection closed [Root exception is java.io.IOException: connection closed``` errors. Most of the changes are for getting the ldap configs from CertLdapLoginModule down to LDAPSearcher.

Ldap lookup is expensive, but it will be done only when roles are not cached. We could further improve performance by introducing a real ldap connection pool (with validation and restoring stale connections).


dcrissman:
I believe https://github.com/lightblue-platform/lightblue-core/pull/402 will fix the compilation issues you are having.

dcrissman:
Merging. This is already included in 1.4.1

PR #156, opened 2015-06-02
update pom

dcrissman:


lightblue-rest-integration-test/pom.xml
```diff
@@ -33,17 +33,21 @@
         <version>1.5.0-SNAPSHOT</version>
         <classifier>classes</classifier>
     </dependency>
+    <dependency>
+        <groupId>junit</groupId>
+        <artifactId>junit</artifactId>
+        <scope>compile</scope><!-- Yes, compile -->
```
paterczm: Why compile?

dcrissman: The integration test module is for testing. So it takes advantage of a couple of junit features.

PR #159, opened 2015-07-12
Add sudo:false to .travis.yml

alechenninger:
This will allow travis to use container infrastructure for significantly reduced build wait time.


dcrissman:
I will merge this as I know this commit is not what caused the problem.

PR #160, opened 2015-07-12
Fixes #300: Default query to objectType:entityName in simple find

alechenninger:
Closes https://github.com/lightblue-platform/lightblue-core/issues/300


alechenninger:
I'd write a unit test for this but it is hard to test lightblue-rest because of lack of dependency injection.

alechenninger:
Actually, I think I could use `RestConfiguration.setFactory`...

bserdar:
This is not the best way to do this. Lets make query optional for find()
operation in core, and then you can pass empty query.

alechenninger:
It seemed like since there were already defaults in the simple find query, it fit neatly here. You cannot pass an empty projection, for instance, but you can in the simple find query.

jewzaam:
Query is now optional: https://github.com/lightblue-platform/lightblue-core/pull/432

PR #163, opened 2015-07-30
Synchronization APIs

bserdar:


jewzaam:
Wanted to note this PR removed deprecated insert REST API's and caused issues for clients when it was released.  We need to take care to not remove functionality in minor releases.

https://github.com/lightblue-platform/lightblue-rest/issues/179

PR #174, opened 2015-09-11
Fix metadata checks; metadata cache uses one extra document, and the …

bserdar:
…test were not counting that


jewzaam:
+1 on review, once build is successful it can be merged

PR #192, opened 2016-01-14
Rename and fixes for upstream changes

dcrissman:
This is technically a breaking change, but there is only a single consumer of the class and that is in lightblue-client. I will create a PR there as well.


dcrissman:
https://github.com/lightblue-platform/lightblue-client/pull/252 must also be merged.

PR #193, opened 2016-01-19
Fixes #162 - Load external resources

### LINKED ISSUE #162: Dynamically configure maven dependencies for lightblue plugins
From a maven dependency perspective, all possible plugins must be listed in lightblue-rest's pom file. You can already fine tune which plugins are actually used at runtime via the lightblue-metadata.json configuration, but all the possible plugins must still be listed in the pom.

It would be idea if you could truly configure what plugins you wanted to use. This would be beneficial for highly customized plugins that not anyone would want to use, or if you just want a lighter weight installation.

dcrissman:
Allows lightblue-rest to load external jar files and directories. It does this by adding a URLClassLoader to the currentThread's class loader.

I also removed the ldap hard requirement as a test.

This is a part of #162.

### DIFF
--- a/common/src/main/java/com/redhat/lightblue/rest/ExternalResourceConfiguration.java
+++ b/common/src/main/java/com/redhat/lightblue/rest/ExternalResourceConfiguration.java
@@ -1,6 +1,10 @@
 package com.redhat.lightblue.rest;
 
+import java.io.File;
+import java.io.FilenameFilter;
 import java.io.IOException;
+import java.net.MalformedURLException;
+import java.net.URL;
 import java.util.Collections;
 import java.util.HashSet;
 import java.util.Set;
@@ -26,9 +30,11 @@ import com.redhat.lightblue.util.JsonInitializable;
  */
 public class ExternalResourceConfiguration implements JsonInitializable {
 
-    private final Set<String> externalPaths = new HashSet<>();
+    public final static String PROTOCOL_FILE = "file://";
 
-    public Set<String> getExternalPaths() {
+    private final Set<URL> externalPaths = new HashSet<>();
+
+    public Set<URL> getExternalUrls() {
         return Collections.unmodifiableSet(externalPaths);
     }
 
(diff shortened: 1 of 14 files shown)


dcrissman:
Ready for review

dcrissman:
This possibly constitutes a breaking change as it removes ldap as a hard dependency. Not a big deal if a deployment doesn't use ldap, but if it does, then ldap will now need to be deployed separately.

common/src/main/java/com/redhat/lightblue/rest/ExternalResourceConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+
+    private final Set<URL> externalPaths = new HashSet<>();
+
+    public Set<URL> getExternalUrls() {
+        return Collections.unmodifiableSet(externalPaths);
+    }
+
+    public ExternalResourceConfiguration() {}
+
+    public ExternalResourceConfiguration(JsonNode node) throws IOException {
```
BVulaj: What's the point of initializing this from a jsonnode?  Maybe some documentation around this if we expect to read in a json file or something.

common/src/main/java/com/redhat/lightblue/rest/ExternalResourceConfiguration.java
```diff
@@ -0,0 +1,112 @@
...
+        if (!file.exists()) {
+            throw new IllegalArgumentException("File does not exist: " + file.getAbsolutePath());
+        }
+
+        if (file.isDirectory()) {
+            File[] files = file.listFiles(new FilenameFilter() {
+
+                @Override
+                public boolean accept(File dir, String name) {
+                    if (name.endsWith(".jar")
```
BVulaj: Given this, won't externalPaths end up with directories as well as jar flies?  Shouldn't this recurse and find them all?

common/src/main/java/com/redhat/lightblue/rest/ExternalResourceConfiguration.java
dcrissman: The class level javadoc includes an example of what the json config should look like.

PR #199, opened 2016-02-01
bump to next major version

dcrissman:
related to https://github.com/lightblue-platform/lightblue-rest/issues/162

The issue changes how plugins are loaded into lightblue, because of that I feel a bump in major version is warranted.

### DIFF
--- a/auth/pom.xml
+++ b/auth/pom.xml
@@ -4,10 +4,9 @@
     <parent>
         <groupId>com.redhat.lightblue.rest</groupId>
         <artifactId>lightblue-rest-pom</artifactId>
-        <version>1.11.0-SNAPSHOT</version>
+        <version>2.0.0-SNAPSHOT</version>
     </parent>
     <artifactId>lightblue-rest-auth</artifactId>
-    <version>1.11.0-SNAPSHOT</version>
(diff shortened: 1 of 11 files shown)


dcrissman:
There is some discussion around the definition of a "breaking change".

bserdar:
My understanding is that a breaking change implies major version update,
but not the other way around. Major version updates don't have to be
breaking changes.

dcrissman:
The change in #162 alters how "plugins" are loaded. While not technically a code breaking change, no longer will "plugins" just be made available in lightblue-rest as a maven dependency. Instead each will need to be individually installed and configured on the host.

If an existing deployment of lightblue were to be upgraded, functionality could be potentially lost. Because of this, I think a major version change would help indicate to the deployment owner that they need to be aware that something big changed. I also created https://github.com/lightblue-platform/lightblue-docs-specs/issues/34 to write documentation for this.

dcrissman:
A point @jewzaam made.

> This isn't breaking to existing deployments unless they had those hooks configured.

This is true. A ClassNotFoundException would be thrown on the server the first time a data or metadata request was processed as that is when the LightblueFactory is lazy loaded.

PR #205, opened 2016-02-09
Prevent plugins from being loaded multiple times

dcrissman:


common/src/main/java/com/redhat/lightblue/rest/RestConfiguration.java
```diff
@@ -74,7 +74,7 @@ public static LightblueFactory getFactory() {
         return factory;
     }
 
-    public static LightblueFactory getFactory(
+    public synchronized static LightblueFactory getFactory(
```
bserdar: Inefficient. Instead:
1) Declare factory as volatile
2) remove synchronized
3) getFactory() {
       LightblueFactory f=factory;
       if(f==null) {
            synchronized(this) {
                    if(factory==null)  initialize factory
                    f=factory;
            }
   return f;
}

dcrissman: > synchronized(this)

That was my initial thought also, but we are in a static context so it won't work. I could create a lock object also in a volatile state, but I guess that seemed silly over just making the method synchronized.

common/src/main/java/com/redhat/lightblue/rest/RestConfiguration.java
```diff
@@ -60,36 +60,40 @@ public static DataSourcesConfiguration getDatasources() {
...
+        LightblueFactory f = factory;
+        if (f == null) {
             return getFactory(loadDefaultPlugins());
         }
-        return factory;
+        return f;
     }
 
-    public static LightblueFactory getFactory(
+    public synchronized static LightblueFactory getFactory(
```
dcrissman: Still an open discussion about synchronized method vs. using a synchronized block.

common/src/main/java/com/redhat/lightblue/rest/RestConfiguration.java
bserdar: You can use synchronize on RestConfiguration.class.

dcrissman:
@bserdar ready for review.

PR #223, opened 2016-04-27
Added a resource and command for re-indexing

BVulaj:


metadata/src/main/java/com/redhat/lightblue/rest/metadata/AbstractMetadataResource.java
```diff
@@ -278,4 +280,12 @@ else if(roles.contains(MetadataConstants.ROLE_ANYONE)){
         throw new SecurityException("Unauthorized Request. One of the following roles is required: " + roles);
     }
 
+    @POST
+    @LZF
+    @Path("/{entity}/{version}/reindex")
```
bserdar: version isn't required

metadata/src/main/java/com/redhat/lightblue/rest/metadata/hystrix/ReIndexCommand.java
```diff
@@ -0,0 +1,47 @@
...
+        }
+    }
+
+
+    /* (non-Javadoc)
+     * @see com.netflix.hystrix.HystrixCommand#run()
+     */
+    @Override
+    protected String run() throws Exception {
+        EntityMetadata emd = getMetadata().getEntityMetadata(entity, version);
```
bserdar: When version is removed, you can use Factory.getCRUDController(ei.getDataStore().getBackend())

PR #228, opened 2016-05-03
Pull in the notification hook

bserdar:


dcrissman:
Approving as a temporary work around until plugins are fixed.

PR #229, opened 2016-05-04
Expose generator API

bserdar:


crud/src/main/java/com/redhat/lightblue/rest/crud/AbstractCrudResource.java
```diff
@@ -259,6 +260,36 @@ public Response bulk(String request) {
...
+     * @param n Number of values to be generated
+     *
+     * @return A lightblue response, with "processed" containing an array of generated values.
+     */
+    @GET
+    @LZF
+    @Path("/generate/{entity}/{version}/{field}")
+    public Response generate(@PathParam("entity") String entity,
+                             @PathParam("version") String version,
+                             @PathParam("field") String field,
```
jewzaam: Is it better to call this path instead of field?

crud/src/main/java/com/redhat/lightblue/rest/crud/hystrix/GenerateCommand.java
```diff
@@ -0,0 +1,131 @@
...
+            Set<String> userRoles=new HashSet<>();
+            if(sreq!=null) {
+                for(String role:mdResolver.getMetadataRoles()) {
+                    if(sreq.isUserInRole(role))
+                        userRoles.add(role);
+                }
+            }
+            LOGGER.debug("user roles:{}",userRoles);
+            EntityMetadata emd=mdResolver.getCompositeMetadata();
+            if(emd.getAccess().getFind().hasAccess(userRoles)||
```
jewzaam: Would remove find from this.  A find operation will never generate a value for something.
Would consider removing update access.  When would an update generate a value?  Maybe for a date if override is true, but that's not the intended use case.  If someone wants a date, they don't need to use a lightblue generator.

crud/src/main/java/com/redhat/lightblue/rest/crud/hystrix/GenerateCommand.java
```diff
@@ -0,0 +1,131 @@
...
+               emd.getAccess().getInsert().hasAccess(userRoles)||
+               emd.getAccess().getUpdate().hasAccess(userRoles)) {
+                LOGGER.debug("User has access, looking up {}",fieldPath);
+                FieldTreeNode ftn=emd.resolve(fieldPath);
+                LOGGER.debug("Field={}",ftn);
+                SimpleField fld=(SimpleField)ftn;
```
jewzaam: What if the field given is not a simple field?  Maybe the request is messed up and path is to an object or array.  Would be better to handle here than bubble up class cast exception.

crud/src/main/java/com/redhat/lightblue/rest/crud/hystrix/GenerateCommand.java
```diff
@@ -0,0 +1,131 @@
...
+                        JsonNode node=GeneratedFields.generate(factory,fld,emd);
+                        arr.add(node);
+                    }
+                    com.redhat.lightblue.Response r=new com.redhat.lightblue.Response();
+                    r.setStatus(OperationStatus.COMPLETE);
+                    r.setEntityData(arr);
+                    return new CallStatus(r);
+                }
+            } else {
+                throw Error.get(CrudConstants.ERR_NO_ACCESS, "insert " + mdResolver.getTopLevelEntityName());
```
jewzaam: /insert/generate/

crud/src/main/java/com/redhat/lightblue/rest/crud/AbstractCrudResource.java
bserdar: Changed that to path

crud/src/test/java/com/redhat/lightblue/rest/crud/ITCaseCrudResourceTest.java
```diff
@@ -317,4 +317,20 @@ public void testLock() throws IOException, ClassNotFoundException, NoSuchMethodE
...
+        RestConfiguration.getFactory().getMetadata().createNewMetadata(em);
+
+        String result=cutCrudResource.generate("generate","1.0.0","number",1).getEntity().toString();
+        System.out.println("Generated:"+result);
+        JSONAssert.assertEquals("{\"processed\":[\"50000000\"]}",result,false);
+
+        result=cutCrudResource.generate("generate","1.0.0","number",3).getEntity().toString();
+        System.out.println("Generated:"+result);
+        JSONAssert.assertEquals("{\"processed\":[\"50000001\",\"50000002\",\"50000003\"]}",result,false);
+    }
```
jewzaam: Test when no version is supplied?

crud/src/main/java/com/redhat/lightblue/rest/crud/hystrix/GenerateCommand.java
bserdar: Update would generate a value if you add an array element containing a
generated field. So i'll check update and insert.

crud/src/main/java/com/redhat/lightblue/rest/crud/hystrix/GenerateCommand.java
bserdar: There's a catch-all at the end with ERR_REST_GENERATE.

crud/src/main/java/com/redhat/lightblue/rest/crud/hystrix/GenerateCommand.java
bserdar: Copy/paste error. Fixed.

crud/src/test/java/com/redhat/lightblue/rest/crud/ITCaseCrudResourceTest.java
bserdar: Modified the test to verify that as well

crud/src/main/java/com/redhat/lightblue/rest/crud/AbstractCrudResource.java
```diff
@@ -275,19 +275,19 @@ public Response bulk(String request) {
     @Path("/generate/{entity}/{version}/{field}")
```
jewzaam: /field/path

bserdar: Fixed.

PR #230, opened 2016-05-04
Fixing ldap cache and removing hystrix from ldap cache command class

jewzaam:
In prod we see intermittent issues getting roles from ldap that result in requests failing for clients.  It's not frequent, but only thing I can see as a problem is we expire records in the fallback cache for ldap roles.  So, this change removes that expiration.  In addition, I ripped hystrix out of that command class to simplify things and make it easier to follow.  We wanted to remove hystrix anyway, so this is a start.

### DIFF
--- a/auth/src/main/java/com/redhat/lightblue/rest/authz/RolesCache.java
+++ b/auth/src/main/java/com/redhat/lightblue/rest/authz/RolesCache.java
@@ -48,19 +48,7 @@ public class RolesCache {
         fallbackRolesCache = CacheBuilder.newBuilder()
                 .concurrencyLevel(10) // handle 10 concurrent request without a problem
                 .maximumSize(500) // Hold 500 sessions before remove them
-                .expireAfterWrite(5, TimeUnit.HOURS) // If the session is inactive for more than 5 hours, remove it
-                .removalListener(
-                        new RemovalListener<String, List<String>>() {
-                            {
-                                LOGGER.debug("Removal Listener created");
-                            }
-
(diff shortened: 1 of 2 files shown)


auth/src/main/java/com/redhat/lightblue/rest/authz/RolesCache.java
```diff
@@ -48,19 +48,7 @@ public void onRemoval(@ParametersAreNonnullByDefault RemovalNotification notific
...
-                                LOGGER.debug("Removal Listener created");
-                            }
-
-                            @Override
-                            public void onRemoval(@ParametersAreNonnullByDefault RemovalNotification notification) {
-                                LOGGER.debug("This data from " + notification.getKey() + " evacuated due:" + notification.getCause());
-                            }
-                        }
-                ).build();
+                .build();
```
bserdar: There is no expiration now? If there isn't load, it'll stay in the cache indefinitely.

jewzaam: Every time roles are fetch successfully from ldap they're updated in both caches:

https://github.com/lightblue-platform/lightblue-rest/blob/master/auth/src/main/java/com/redhat/lightblue/rest/authz/RolesCache.java#L67-L71

So yes the fallback will not expire but it will kept updated.  If a client is idle for a period of time and then ldap search fails, would it not be better to use what we knew in the past?  The cache has a size limit so it can't grow beyond 500, but that's the only limit.

PR #243, opened 2016-06-03
Fix #242: add diff api

### LINKED ISSUE #242: Metadata diff API?
https://github.com/lightblue-platform/lightblue-core/issues/260

 jblashka commented on Dec 15, 2014

Not sure where this should go, but it would be nice if there was metadata diff api available that would return the diff of two versions of an entity's metadata, instead of having to download the two metadatas and find some comparison tool.

Maybe /rest/metadata/terms/1.0.0/1.0.1 ?

bserdar:


metadata/src/main/java/com/redhat/lightblue/rest/metadata/cmd/DiffCommand.java
```diff
@@ -0,0 +1,100 @@
...
+    public String run() {
+        LOGGER.debug("run: entity={}, version1={}, version2={}", entity, version1, version2);
+        Error.reset();
+        Error.push(getClass().getSimpleName());
+        try {
+            EntityMetadata md1 = getMetadata().getEntityMetadata(entity, version1);
+            EntityMetadata md2 = getMetadata().getEntityMetadata(entity, version2);
+            if (md1 != null) {
+                JSONMetadataParser parser = getJSONParser();
+                JsonNode jmd1=parser.convert(md1);
```
dcrissman: nit pick: could just perform this convert after the md2 != null check. No point in doing the conversion work if md2 is null.

dcrissman:
Would be nice to see a couple unit tests for this.

bserdar:
Done

PR #247, opened 2016-09-21
Add ability to extract uid from cert principal for LDAP lookup name

derek63:


derek63:
@dcrissman got any time to look at this?

PR #248, opened 2016-09-22
Add ability to validate location from client certificate

derek63:
against configured login module environment parameter


derek63:
@dcrissman Would you mind taking a look at this one too?  Thanks

PR #250, opened 2016-09-30
Fixes #249 - Loggingfilter

### LINKED ISSUE #249: Include a session UUID in MDC
This would allow us to group all logs that occurred within a session together. The logger pattern would also need to expose this.

dcrissman:
Fixes https://github.com/lightblue-platform/lightblue-rest/issues/249

Add LoggingFilter to create a UUID for every request. That UUID is registered with the MDC so that it can be included in log4j patterns using %X{RequestUUID}


common/src/main/java/com/redhat/lightblue/rest/logging/LoggingFilter.java
```diff
@@ -0,0 +1,65 @@
...
+
+    @Override
+    public void destroy() {
+        //Do Nothing!
+    }
+
+    @Override
+    public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain) throws IOException, ServletException {
+        try {
+            String requestUUID = UUID.randomUUID().toString();
```
bserdar: This returns a rather large hard to read garbage. Instead, why not have a static AtomicInteger with addAndGet()? That might cause problems with ,multiple hosts, so maybe a static hostname variable that is initialized once, concatenated with that int? So it becomes lightbluecrud1-1234, which might be more human readable.

dcrissman: The idea is to have a unique identifier for all logs produced within a session for all time. For example, if I knew the uuid for an exception I could grep all the logs for that uuid for all logs associated with the exception and hopefully get a more complete picture of what happened.

From what I can tell, your suggestion would achieve that goal (although we might want to make it volatile also) so long as the application isn't restarted. At which point the Integer starts over. So when grep is performed on the logs now, you could get multiple connection log sets. Is there a way to make AtomicInteger unique across restarts? I suppose we could persist the value to disk, but that seems like a lot of work compared to just using a uuid.

paterczm: Or just make requestUUID shorter, say 6 characters.

dcrissman: > maybe a static hostname

From a splunk perspective, it does keep track of which host the log came from. But obviously that is non-helpful for non-splunk users.

bserdar: UUID.getLeastSignificantBits should be sufficient I think, so that at least makes it smaller.

bserdar:
LSB would return the last part of the string, MSB would return the first
part. I suppose both should be somewhat equally distinct.

dcrissman:
> LSB would return the last part of the string, MSB would return the first part. I suppose both should be somewhat equally distinct.

I am honestly good either way. Whichever then you'd like.

PR #251, opened 2016-10-05
Include principal as MDC value

dcrissman:
Fixes https://github.com/lightblue-platform/lightblue-core/issues/697


dcrissman:
@bserdar - Have a few moments to look at this one? It is essentially a continuation and enhancement of https://github.com/lightblue-platform/lightblue-rest/pull/250

common/src/main/java/com/redhat/lightblue/rest/logging/LoggingFilter.java
```diff
@@ -51,9 +55,21 @@ public void doFilter(ServletRequest req, ServletResponse resp, FilterChain chain
...
+            LOGGER.debug("log for testing");
+
             chain.doFilter(req, resp);
         } finally {
             MDC.remove(HEADER_REQUEST_UUID);
+            MDC.remove(HEADER_REQUEST_PRINCIPAL);
+        }
+    }
+
+    private void principal(HttpServletRequest httpReq) {
```
bserdar: Ugly name. Also, would've been easier to read if this was inline rather than a function I think.

dcrissman:
Threw in another change to make the RequestUUID just RequestID as that better matches what it is now.

PR #256, opened 2016-10-25
Add TLS support for LDAP

derek63:


auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProvider.java
```diff
@@ -43,10 +40,10 @@
     InitialLdapContextProvider ldapContextProvider;
     String ldapSearchBase;
 
-    public LightblueLdapRoleProvider(String server, String searchBase, String bindDn, String bindDNPwd) throws NamingException {
+    public LightblueLdapRoleProvider(String server, String searchBase, String bindDn, String bindDNPwd, Boolean useSSL, String trustStore, String trustStorePassword) throws NamingException {
```
dcrissman: Would there be value in keeping both constructors? You wouldn't need to touch LightblueLdapRoleProviderTest then.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/InitialLdapContextProvider.java
```diff
@@ -21,7 +20,7 @@
 
     private Hashtable<String, Object> env = new Hashtable<>();
 
-    public InitialLdapContextProvider(String server, String bindDn, String bindDNPwd) {
+    public InitialLdapContextProvider(String server, String bindDn, String bindDNPwd, Boolean useSSL, String trustStore, String trustStorePassword) {
```
dcrissman: Would there be value in keeping both constructors?

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/InitialLdapContextProvider.java
```diff
@@ -33,6 +32,11 @@ public InitialLdapContextProvider(String server, String bindDn, String bindDNPwd
         }
         env.put(Context.INITIAL_CONTEXT_FACTORY, "com.sun.jndi.ldap.LdapCtxFactory");
         env.put(Context.PROVIDER_URL, server);
+        if(useSSL) {
+            System.setProperty("javax.net.ssl.trustStore", trustStore);
+            System.setProperty("javax.net.ssl.trustStorePassword", trustStorePassword);
```
dcrissman: Where/When do these system properties get read back in?

derek63: The built-in Java LDAP support uses these when you connect to an ldaps:// endpoint.  Using System.setProperty is totally tacky, which is why I opened https://github.com/lightblue-platform/lightblue-rest/issues/257.  However, we don't have time for all that at the moment...

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProvider.java
derek63: Perhaps, let me add the default one back, and revert the LightblueLdapRoleProviderTest change, and we can revisit this when we get to changing out the LDAP library (see https://github.com/lightblue-platform/lightblue-rest/issues/257)

PR #260, opened 2016-11-02
Fix LDAP connection memory leak

derek63:


derek63:
@paterczm Please review/merge deploy to see if this fixes the issue we've been experiencing

PR #261, opened 2016-11-02
Turn LdapRepository into a singleton to maintain a single ldap connection pool

paterczm:
This is a follow up to https://github.com/lightblue-platform/lightblue-rest/pull/260.

@derek63, the connection pool is initialized each time LdapRepository.initialize() is called, which is every time results are not found in cache. With your fix applied, I think we will end up having as many pools as we have different users calling within roles cache expiry threshold. We need to ensure we have only one pool at all times or not use a pool at all.

Here is my quick fix. I don't like it very much, because it makes it impossible to create multiple LdapRepository instances with different LdapConfigurations. We don't need to do that right now, so I guess that's fine.

### DIFF
--- a/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapRepository.java
+++ b/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapRepository.java
@@ -21,9 +21,19 @@ public class LdapRepository {
     private static final Logger LOGGER = LoggerFactory.getLogger(LdapRepository.class);
 
     LdapConfiguration ldapConfiguration;
-    LDAPConnectionPool connectionPool;
+    // Connection pool is a singleton
+    static LDAPConnectionPool connectionPool;
 
-    public LdapRepository(LdapConfiguration ldapConfiguration) {
+    /**
+     * Subsequent LdapRepository initializations will ignore LdapConfiguration, because connection pool is already initialized.
+     *
+     * @param ldapConfiguration
+     */
+    public static LdapRepository getInstance(LdapConfiguration ldapConfiguration) {
+        return new LdapRepository(ldapConfiguration);
+    }
+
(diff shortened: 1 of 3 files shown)


derek63:
LdapRepository.initialize() only gets called when the [connection pool is null](https://github.com/lightblue-platform/lightblue-rest/blob/master/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapRepository.java#L72), which only happens when a new LdapRepository is created and before it is used.  After the change I made last night, [only one LdapRepository is ever created](https://github.com/lightblue-platform/lightblue-rest/blob/master/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LightblueLdapRoleProvider.java#L70) when LightblueLdapRoleProvider is created.  That LdapRepository is [passed to each instance of CachedLdapFindUserRolesByUidCommand ](https://github.com/lightblue-platform/lightblue-rest/blob/master/auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/CachedLdapFindUserRolesByUidCommand.java#L47).  With the last PR, this code implementing the same logic for connection/instance management that was in place when we were using ContextProvider (which was the intent originally as well, despite the execution).

So, merging the previous PR ensured there is only one connection pool ever created.  That said, this change makes that even more explicit, and helps to prevent usage mistakes like I made in the initial implementation, so thanks for getting this changed.  

If we're still experiencing problems even after this change, we can just take out the connection pool altogether, as we have been getting along fine up until now without one.

paterczm:
I tried this with a debugger - LdapRepository is constructed on every hit to Lightblue, just not initialized if roles are found in cache. Noticed something is off when class histogram was still showing multiple pool instances even after #260 was deployed.

derek63:
Okay, well sorry I didn't get it right on the first two go-arounds.  Anyway, it seems like things are good since deploying this change, but we'll keep a close watch on it to confirm.  Thanks for your help debugging and fixing.

PR #270, opened 2016-12-07
Rewrote authnz

paterczm:
Rewrote the auth module to make it manageable again. This is a pre-requirement to moving this logic out of lightblue-rest into a standalone library.


auth/src/main/java/com/redhat/lightblue/rest/auth/jboss/CertLdapLoginModule.java
```diff
@@ -127,7 +130,18 @@ public void initializeLightblueLdapRoleProvider() throws NamingException {
             ldapConf.poolMaxConnectionAgeMS(Integer.parseInt((String)options.get(POOL_MAX_CONNECTION_AGE_MS)));
         }
 
-        lbLdap = new LightblueLdapRoleProvider(searchBase, ldapConf);
+        int rolesCacheExpiry = 5*60*1000; // default 5 minutes
+        if (options.containsKey(ROLES_CACHE_EXPIRY_MS)) {
```
paterczm: It's decoupled from puppet changes this way. We don't need to update puppet unless we want to change the optional values.

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapRolesProvider.java
```diff
@@ -95,29 +140,54 @@ private void initialize() throws Exception {
...
+        String filter = "(uid=" + username + ")";
+
+        SearchRequest searchRequest = new SearchRequest(searchBase, SearchScope.SUB, filter);
+        SearchResult searchResult = connectionPool.search(searchRequest);
+        List<SearchResultEntry> searchResultEntries = searchResult.getSearchEntries();
+
+        if(searchResultEntries.isEmpty()) {
+            LOGGER.warn("No result found roles for user: " + username);
+            return new HashSet<String>();
+        } else if (searchResultEntries.size() > 1) {
```
paterczm: I don't even know if it's possible to happen (a uid= query matching more than one user), but if it can, it seems like an error. Authorization is sensitive, I don't want to introduce ambiguity by merging roles from all users who match.

derek63: It is possible if they are at different locations in the directory, but I agree its probably best left as is in this case

paterczm: Different locations with the same base?

PR #271, opened 2016-12-12
Add ability to use locking API with POST request

derek63:


derek63:
@bserdar Can you review when you have the chance?  Thx

crud/src/main/java/com/redhat/lightblue/rest/crud/LockRequest.java
```diff
@@ -0,0 +1,64 @@
+package com.redhat.lightblue.rest.crud;
+
+public class LockRequest {
+    String operation;
```
bserdar: Add constants for operation pls.

PR #276, opened 2016-12-13
Don't use double checked locking for CertLdapLoginModule initialization

paterczm:
https://en.wikipedia.org/wiki/Double-checked_locking

Locking on each getRoleSets call (every Lightblue request). Hope it does not affect performance.

LdapConfiguration will be initialized only once - this will improve performance.

### DIFF
--- a/auth/src/main/java/com/redhat/lightblue/rest/auth/jboss/CertLdapLoginModule.java
+++ b/auth/src/main/java/com/redhat/lightblue/rest/auth/jboss/CertLdapLoginModule.java
@@ -84,7 +84,7 @@ public class CertLdapLoginModule extends BaseCertLoginModule {
     private static String environment;
 
     // LightblueRoleProvider singleton
-    private static RolesProvider lbLdap = null;
+    private static volatile RolesProvider lbLdap = null;
 
     public CertLdapLoginModule() {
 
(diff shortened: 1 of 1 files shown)


auth/src/main/java/com/redhat/lightblue/rest/auth/jboss/CertLdapLoginModule.java
```diff
@@ -99,49 +99,51 @@ public void initialize(Subject subject, CallbackHandler callbackHandler, Map<Str
...
-        }
-        if (options.containsKey(POOL_MAX_CONNECTION_AGE_MS)) {
-            ldapConf.poolMaxConnectionAgeMS(Integer.parseInt((String)options.get(POOL_MAX_CONNECTION_AGE_MS)));
-        }
-
-        int rolesCacheExpiry = 5*60*1000; // default 5 minutes
-        if (options.containsKey(ROLES_CACHE_EXPIRY_MS)) {
-            rolesCacheExpiry = Integer.parseInt((String)options.get(ROLES_CACHE_EXPIRY_MS));
-        }
+        synchronized(LdapRolesProvider.class) {
```
bserdar: This will synchronize every time initialize is called. Why not use volatile?

paterczm: > Volatile doesn't mean what you think, either
> A commonly suggested nonfix is to declare the resource field of SomeClass as volatile. However, while the JMM prevents writes to volatile variables from being reordered with respect to one another and ensures that they are flushed to main memory immediately, it still permits reads and writes of volatile variables to be reordered with respect to nonvolatile reads and writes. That means -- unless all Resource fields are volatile as well -- thread B can still perceive the constructor's effect as happening after resource is set to reference the newly created Resource.

(from http://www.javaworld.com/article/2074979/java-concurrency/double-checked-locking--clever--but-broken.html)

If I understand this correctly (and unless something has changed, it's a really old article), volatile will not affect the properties of LdapRolesProvider, unless they are volatile as well.

There is also https://en.wikipedia.org/wiki/Initialization-on-demand_holder_idiom, but it only works for initialization which cannot fail, so not good.

bserdar:
afaik, volatile is a memory barrier. A write to a volatile variable will
force all waiting writes to be flushed, so it should be safe to use
volatile.

paterczm:
That article from 2001 probably speaks of volatile implementation prior to jdk5. So I'm following your recommendation.

PR #278, opened 2017-01-24
Saved searches

bserdar:


crud/src/main/java/com/redhat/lightblue/rest/crud/AbstractCrudResource.java
```diff
@@ -78,6 +90,108 @@
         java.security.Security.setProperty("networkaddress.cache.ttl", "30");
     }
 
+    @GET
+    @Path("/search/{searchName}/{entity}")
+    public Response runSavedSearch(@PathParam("searchName") String searchName,
```
bvulaj: Is there someway to get a list of available searches?

bvulaj: Are searches specific to an entity?  Or could getById apply to multiple entities?

bserdar:
Saved searches are specific to an entity. There can be a getById for user,
and getById for product, different searches.

bserdar:
Saved searches are first-class entities. Search the savedSearch entity.

bvulaj:
@bserdar so how would that work, then?

/search/getById/user?
/search/getById/product?

What if I wanted to see available searches for user?  Do I have to write an lb query to filter?

I guess what I am getting at is that it might make more sense to users to expose the searches underneath their respective entity.

/search/user/getById

bserdar:
I agree, /search/user/getById is better. Pushed a change for that.

Should I add APIs to get list of searches, etc? Currently you can do that
with:

/find/savedSearch?Q=entity:user

bvulaj:
I think it would be nice if `/search/user` returned a list of searches.  It's a convenience method, but I think it makes sense.

crud/src/main/java/com/redhat/lightblue/rest/crud/AbstractCrudResource.java
```diff
@@ -78,6 +93,156 @@
         java.security.Security.setProperty("networkaddress.cache.ttl", "30");
     }
 
+    @GET
+    @LZF
+    @Path("/searches/{entity}")
```
bvulaj: I think I would just keep this as `search` rather than change verbiage.  I think that makes sense.  It shouldn't clash with any of the actual search methods.

bserdar:
This doesn't, but /searches/entity/version does. That's why I changed that
to searches. Or, I can remove the API with the version, and keep this as
/search/{entity}

bserdar:
Pushed.

PR #288, opened 2017-06-08
WIP: Added metrics health check for CRUD controller and health check servlet endpoint

sameer31:
@alechenninger, @svkm102, @derek63, Kindly review the changes


common/src/main/java/com/redhat/lightblue/rest/health/LightblueHealthCheckServletContextListener.java
```diff
@@ -0,0 +1,38 @@
...
+	static {
+
+		LightblueFactory factory = RestConfiguration.getFactory();
+		CRUDController[] crudControllers = null;
+		try {
+			crudControllers = factory.getFactory().getCRUDControllers();
+		} catch (Exception e) {
+		}
+
+		for (CRUDController crudController : crudControllers) {
```
macgregor: would this cause a NPE?

crud/src/test/java/com/redhat/lightblue/rest/crud/TestHealthCheckServlet.java
```diff
@@ -0,0 +1,59 @@
...
+import org.jboss.shrinkwrap.api.spec.WebArchive;
+import org.jboss.shrinkwrap.resolver.api.maven.Maven;
+import org.junit.Test;
+import org.junit.runner.RunWith;
+
+import com.gargoylesoftware.htmlunit.WebClient;
+import com.gargoylesoftware.htmlunit.html.HtmlPage;
+
+@RunWith(Arquillian.class)
+public class TestHealthCheckServlet {
```
alechenninger: I'm working on setting up the arquillian tests to use a servlet container instead of just CDI (which is what the current classpath is set up to do, and why this won't work).

Once that is set up, we should use the other test that already uses Arquillian to avoid the time expense of multiple test classes each setting up a container / forking them JVM and such. We can chat more about this in scrum tomorrow.

alechenninger:
@sameer31 See https://github.com/sameer31/lightblue-rest/pull/1 – you can merge this to get Arquillian working

crud/src/test/java/com/redhat/lightblue/rest/crud/ITCaseCrudResourceTest.java
```diff
@@ -120,10 +140,19 @@ public void onProcessed() {
...
+                .processOutput(new ProcessOutput(mongodOutput, mongodError, commandsOutput))
+                .artifactStore(new ExtractedArtifactStoreBuilder()
+                    .defaults(mongoD)
+                    .download(new DownloadConfigBuilder()
+                        .defaultsForCommand(mongoD)
+                        .timeoutConfig(new TimeoutConfigBuilder()
+                            .connectionTimeout((int) Duration.ofMinutes(1).toMillis())
+                            .readTimeout((int) Duration.ofMinutes(5).toMillis())
+                            .build())))
+                .build();
```
svkm102: nitpick: align these for better readability

crud/src/test/java/com/redhat/lightblue/rest/crud/ITCaseCrudResourceTest.java
```diff
@@ -189,36 +218,46 @@ public static void clearDatabase() {
...
+        WebArchive archive = ShrinkWrap.create(WebArchive.class, "lightblue.war")
+            .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml")
+            .addAsWebInfResource(Assets.forDocument(CrudWebXmls.forNonEE6Container(RestApplication.class)), "web.xml")
+            .addAsLibraries(Maven.configureResolver()
+                    .workOffline()
+                    .loadPomFromFile("pom.xml")
+                    .resolve("org.jboss.weld.servlet:weld-servlet")
+                    .withTransitivity()
+                    .asFile())
+            .addPackages(true, "com.redhat.lightblue")
```
svkm102: nitpick: align these for better readability

metadata/src/test/java/com/redhat/lightblue/rest/metadata/ITCaseMetadataResourceTest.java
```diff
@@ -74,19 +76,26 @@ public void teardown() {
...
+        WebArchive archive = ShrinkWrap.create(WebArchive.class, "lightblue.war")
                 .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml")
+                .addAsWebInfResource(Assets.forDocument(CrudWebXmls.forNonEE6Container(RestApplication.class)), "web.xml")
+                .addAsLibraries(Maven.configureResolver()
+                        .workOffline()
+                        .loadPomFromFile("pom.xml")
+                        .resolve("org.jboss.weld.servlet:weld-servlet")
+                        .withTransitivity()
+                        .asFile())
+                .addPackages(true, "com.redhat.lightblue")
```
svkm102: nitpick: align these for better readability

metadata/src/test/java/com/redhat/lightblue/rest/metadata/ITMongoIndexManipulationTest.java
```diff
@@ -64,19 +66,26 @@ public void teardown() {
...
+        WebArchive archive = ShrinkWrap.create(WebArchive.class, "lightblue.war")
                 .addAsWebInfResource(EmptyAsset.INSTANCE, "beans.xml")
+                .addAsWebInfResource(Assets.forDocument(CrudWebXmls.forNonEE6Container(RestApplication.class)), "web.xml")
+                .addAsLibraries(Maven.configureResolver()
+                        .workOffline()
+                        .loadPomFromFile("pom.xml")
+                        .resolve("org.jboss.weld.servlet:weld-servlet")
+                        .withTransitivity()
+                        .asFile())
+                .addPackages(true, "com.redhat.lightblue")
```
svkm102: nitpick: align these for better readability

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapRolesProvider.java
```diff
@@ -190,4 +191,25 @@ private void initialize() throws Exception {
         return roles;
     }
 
+    @Override
+    public LdapAuthHealth checkHealth() {
+
+        boolean isHealthy = true;
+        String details = null;
+        try {
+            LDAPConnection ldapConnection = connectionPool.getConnection();
```
alechenninger: Look at `connectionPool.getHealthCheck()`–there are more advanced health checks we can delegate to there.

sameer31: I checked the java doc for the API method in LDAPConnectionPoolHealthCheck class. These methods are anyway invoked when we call connectionPool.getConnection() or when a connection is released. A successful connection instance from pool is only returned if associated health checks pass otherwise it throws LDAPException. 

Do you suggest to make an explicit check on these methods again in our health check?

auth/src/test/java/com/redhat/lightblue/rest/auth/ldap/LdapRoleProviderIntegrationTest.java
```diff
@@ -102,4 +103,11 @@ public void testUserWhoDoesNotExist() throws Exception {
         assertTrue(provider.getUserRoles("idontexist").isEmpty());
     }
 
+    @Test
+    public void testHealthCheck() throws Exception {
+        
+        RolesProviderHealth health = provider.checkHealth();
+        System.out.println(health.details());
+        assertTrue(health.isHealthy());
+    }
```
alechenninger: Is it possible to take the in memory ldap server down and do the health check to see it's unhealthy? That'd be pretty rad :-).

crud-cert-auth/src/main/java/com/redhat/lightblue/rest/crud/cert/auth/health/CrudCertAuthHealthCheckServletContextListener.java
```diff
@@ -0,0 +1,41 @@
...
+    protected HealthCheckRegistry getHealthCheckRegistry() {
+
+        HealthCheckRegistry healthCheckRegistry = new HealthCheckRegistry();
+
+        LightblueFactory factory = RestConfiguration.getFactory();
+        CRUDController[] crudControllers = null;
+        try {
+            crudControllers = factory.getFactory().getCRUDControllers();
+        } catch (Exception e) {
+        }
```
alechenninger: We should probably log a warn or error here if this actually happens, or even just fail.

crud-cert-auth/src/main/webapp/WEB-INF/web.xml
```diff
@@ -76,5 +71,19 @@ xsi:schemaLocation="http://java.sun.com/xml/ns/javaee http://java.sun.com/xml/ns
...
+      com.redhat.lightblue.rest.crud.cert.auth.health.CrudCertAuthHealthCheckServletContextListener
+    </listener-class>
+  </listener>
+  <servlet>
+    <servlet-name>healthCheck</servlet-name>
+    <servlet-class>com.codahale.metrics.servlets.HealthCheckServlet</servlet-class>
+  </servlet>
+  <servlet-mapping>
+    <servlet-name>healthCheck</servlet-name>
+    <url-pattern>/healthcheck</url-pattern>
```
alechenninger: Maybe /v2/health or /health/v2 or /health/complete or something like that? @derek63 what do you think?

Otherwise I feel like "what's the difference between /health and /healthcheck"? This new health endpoint really supersedes the old one.

crud-cert-auth/src/test/java/com/redhat/lightblue/rest/crud/cert/auth/ITCaseCertAuthTest.java
```diff
@@ -0,0 +1,116 @@
...
+import org.junit.runner.RunWith;
+
+import com.fasterxml.jackson.databind.ObjectMapper;
+import com.fasterxml.jackson.databind.node.ObjectNode;
+import com.redhat.lightblue.rest.test.RestConfigurationRule;
+import com.redhat.lightblue.rest.test.support.Assets;
+import com.redhat.lightblue.rest.test.support.CrudWebXmls;
+
+// @RunWith(Arquillian.class)
+public class ITCaseCertAuthTest {
```
alechenninger: Does it work? You could probably exclude the healthcheck endpoint from security constraints and maybe then it will work? (Won't require a cert)

auth/src/main/java/com/redhat/lightblue/rest/auth/CachedRolesProvider.java
```diff
@@ -72,4 +74,9 @@ public CachedRolesProvider(RolesProvider rolesProvider, RolesCache rolesCache) {
         return roles;
     }
 
+    @Override
+    public RolesProviderHealth checkHealth() {
+        return rolesProvider.checkHealth();
+    }
```
alechenninger: Maybe eventually we need health check to not be so binary but instead more like "healthy, degraded, or unhealthy." Seems complex. Hrm!

auth/src/main/java/com/redhat/lightblue/rest/auth/ldap/LdapRolesProvider.java
```diff
@@ -190,4 +191,26 @@ private void initialize() throws Exception {
...
+
+        boolean isHealthy = true;
+        String details = null;
+        
+        try {
+            LDAPConnection ldapConnection = connectionPool.getConnection();
+
+            if (!ldapConnection.isConnected()) {
+                isHealthy = false;
+            }
```
alechenninger: Actually, since the default health check implementation may not do anything: I'd do both. First check `isConnected`, then call `ensureConnectionValidForContinuedUse()`. If both pass, we're healthy.

auth/src/test/java/com/redhat/lightblue/rest/auth/CachedRolesProviderTest.java
```diff
@@ -75,4 +76,26 @@ public void testFallBackRolesCache() throws Exception {
...
+        
+        System.out.println(cachedRolesProvider.checkHealth().details());        
+        Assert.assertTrue(cachedRolesProvider.checkHealth().isHealthy());
+
+    }
+    
+    @Test
+    public void isUnhealthyIfDelegateRolesProviderIsUnhealthy() throws Exception {
+
+        // Mock healthy
```
sameer31: This was a miss due to copy and paste. I have updated it

crud-cert-auth/src/main/webapp/WEB-INF/web.xml
sameer31: We need to see if we would like to have /health as it is just pointing to health.jsp (which contains only project version)?

auth/src/main/java/com/redhat/lightblue/rest/auth/CachedRolesProvider.java
derek63: `If LDAP goes down, we don't necessarily want to bring all lightblue nodes out of rotation`

As you mention, the only roles in the cache are ones that have already been retrieved from LDAP when requests were made (we don't pre-fetch all users/roles in LDAP).  While we don't have a high amount of user/role churn, just because the cache exists does not mean that requests won't start failing if users not in the cache try and make requests.  

I do think we'll need the checks to account for a degraded status at some point as well, in this case, degraded would be when the connection to LDAP was not available, but the roles cache was still serving requests.  As soon as the cache entries expire, a broken connection to LDAP would then change the status to "unhealthy".  Not sure if we know when role cache expiry happens though to do something about it.

Similarly, degraded could be used to communicate when things are returning in their normal timeframes (i.e. LDAP/Mongo requests are slow).  Of course that implies that we have to know how we define slow for whatever operation (which we'll start to learn as we gather more metrics).

auth/src/test/java/com/redhat/lightblue/rest/auth/ldap/LdapRoleProviderIntegrationTest.java
derek63: You can maybe do what you did in the Mongo one and try to connect on a different port than the one being used by in-memory LDAP server

crud-cert-auth/src/main/webapp/WEB-INF/web.xml
derek63: I think this new health check should be reachable on /health (without any v2 or complete adornments), and we should modify existing health.jsp to be version.jsp and serve it up at /version, as knowing the version deployed is still useful.  Also if we keep it at /health, we won't have to change our existing load balancer configurations to hist the new health check, they will start doing it as soon as we deploy this version  :-)

PR #301, opened 2017-08-18
Health check enhancements

derek63:


crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckHealthCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    public CheckHealthCommand(Mediator mediator, HealthCheckRegistry healthCheckRegistry, RequestMetrics metrics) {
+        super(mediator);
+        this.healthCheckRegistry = healthCheckRegistry;
+        this.metrics = metrics;
+    }
+
+    @Override
+    public CallStatus run() {
+
+        RequestMetrics.Context context = metrics.startHealthRequest();
```
alechenninger: I agree it's worth the small cost to capture these metrics. It is worth remembering I think though that these will run async, so this won't actually time how long it takes to run the health checks themselves, even though that may still be useful information.

This makes me think two things:

1. CRUDHealth is probably a little _too_ basic. The service health we are doing internally has arbitrary map (key/value) parameters possibly, matching dropwizard's capabilities (in other words, dropwizard's health checks can add these parameters easily from the CRUDHealth results for example). Then we could neatly add like, what mongo configuration was used and stuff, instead of cramming it all into one string field.
2. In that same vein, CRUDHealth could add a duration field if we wanted (thus getting that more useful timing information). It wouldn't be in Graphite, but it would at least be a part of the diagnostics response for someone to take a peak at.

That said, I don't necessarily think either of those things are essential, though they may be relatively easy to add.

If we wanted to get fancy we could probably add real JMX metric tracking to the health checks, too, maybe by passing the RequestMetrics instance to the getHealthRegistry methods. Anyway, it's a bit extra, so use your judgement of course :-).

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckConfiguration.java
```diff
@@ -21,18 +13,28 @@
...
+
+public class CrudCheckConfiguration {
 
-    private final Logger LOGGER = LoggerFactory.getLogger(CrudCertAuthHealthCheckServletContextListener.class);
-    
-    @Override
-    protected HealthCheckRegistry getHealthCheckRegistry() {
+    private static final Logger LOGGER = LoggerFactory.getLogger(CrudCheckConfiguration.class);
 
+    public static HealthCheckRegistry getHealthCheckRegistry() {
```
alechenninger: Maybe `newHealthCheckRegistry()` or `createHealthCheckRegistry()` makes it clear this doesn't return a singleton.

crud/src/main/java/com/redhat/lightblue/rest/crud/AbstractCrudResource.java
```diff
@@ -93,6 +98,38 @@
...
+    /**
+     * Thread-safe, shared static instance for all requests. Registry is also static, so does not
+     * make much difference to inject. If you wanted to inject, you could do so rather easily using
+     * CDI injection.
+     */
+    private static final RequestMetrics metrics =
+            new DropwizardRequestMetrics(MetricRegistryFactory.getMetricRegistry());
+
+    private static final HealthCheckRegistry healthCheckRegistry =
+            CrudCheckRegistry.getHealthCheckRegistry();
```
alechenninger: Since this is static, I think you should be able to just use the `CrudCheckConfiguration` here. If I understand correctly it should still only create one instance per classloader and it removes the need for the extra class.

alechenninger: Also there is a RestConfiguration class I guess. Maybe all of this static stuff should go there. Not a big deal.

crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckDiagnosticsCommand.java
```diff
@@ -0,0 +1,92 @@
...
+    private static final Logger LOGGER = LoggerFactory.getLogger(CheckDiagnosticsCommand.class);
+
+    private HealthCheckRegistry healthCheckRegistry;
+    private RequestMetrics metrics;
+
+    public CheckDiagnosticsCommand(HealthCheckRegistry healthCheckRegistry, RequestMetrics metrics) {
+        this(null, healthCheckRegistry, metrics);
+    }
+
+    public CheckDiagnosticsCommand(Mediator mediator, HealthCheckRegistry healthCheckRegistry, RequestMetrics metrics) {
```
alechenninger: I think we can probably omit the Mediator like the lock commands do

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckConfiguration.java
```diff
@@ -87,12 +90,12 @@ private RolesProvider createRolesProvider() throws Exception {
 
         LdapConfiguration ldapConfiguration = new LdapConfiguration();
         ldapConfiguration.initializeFromJson(root);
-        
+
         return new CachedRolesProvider(new LdapRolesProvider(searchBase, ldapConfiguration),
```
alechenninger: Ah, I guess we can't get this from the `CertLdapLoginModule.getRolesProvider()` because that's in a separate class loader? Have we tried making that method initialize the provider if it isn't initialized yet? Perhaps even if we can't share the same role provider instance, we can still reuse the initialization code.

/me curses app servers (yet again)

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckResponse.java
```diff
@@ -0,0 +1,32 @@
...
+    }
+
+    @Override
+    public JsonNode toJson() {
+        ObjectNode healthCheckResponse = factory.objectNode();
+        for (Map.Entry<String, HealthCheck.Result> result : healthCheckResults.entrySet()) {
+            ObjectNode resultResponse = factory.objectNode();
+            resultResponse.put("healthy", result.getValue().isHealthy());
+            resultResponse.put("message", result.getValue().getMessage());
+            healthCheckResponse.put(result.getKey(), resultResponse);
```
alechenninger: the Results can have arbitrary key/value pairs, too, which I think would be nice to take advantage of.

crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckHealthCommand.java
```diff
@@ -0,0 +1,92 @@
...
+        LOGGER.debug("run: health");
+        Error.reset();
+        Error.push("rest");
+        Error.push(getClass().getSimpleName());
+        try {
+            Map<String, HealthCheck.Result> healthCheckResults = new LinkedHashMap<>();
+            CallStatus callStatus = new CallStatus();
+            for (Map.Entry<String, Result> entry : CrudCheckRegistry.getHealthCheckRegistry().runHealthChecks().entrySet()) {
```
derek63: Fixed to use local HealthCheckRegistry instance, I think we should keep health and diagnostic command separate for now, as it helps to reinforce the logical difference between the two, and probably won't change all that much (maybe adding new HealthChecks and registry entries).

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckConfiguration.java
derek63: Renamed to createHealthCheckRegistry and createDiagnosticCheckRegistry

crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckHealthCommand.java
derek63: Yeah I keep forgetting that those checks are running async...I'll leave out the duration addition for now, but we can add later if we find ourselves looking for it.

Not sure if I understand what you mean by "adding real JMX metric tracking", don't I get that with the metrics instance that is set locally here (passed in from AbstractCrudResoure)?

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckConfiguration.java
derek63: Yeah unfortunately the "login modules run in a separate classloader" issue is why this config file needs to exist (we tried without it in the beginning couldn't make it work).  Not sure we can share any initialization code, as the instance used by the login module is read from config parameters read from standalone.xml, not from the nice JSON file like we have.  App servers strike again

crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckDiagnosticsCommand.java
derek63: Removed

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckResponse.java
derek63: Added details key-value map to JSON

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckRegistry.java
```diff
@@ -17,15 +17,15 @@ private CrudCheckRegistry() {
...
     }
 
 
     public static HealthCheckRegistry getHealthCheckRegistry() {
         if(healthCheckRegistry == null) {
-            healthCheckRegistry = CrudCheckConfiguration.getHealthCheckRegistry();
+            healthCheckRegistry = CrudCheckConfiguration.createHealthCheckRegistry();
         }
         return healthCheckRegistry;
     }
```
alechenninger: Do we still need both CrudCheckRegistry and CrudCheckConfiguration?

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckResponse.java
```diff
@@ -24,6 +24,13 @@ public JsonNode toJson() {
             ObjectNode resultResponse = factory.objectNode();
             resultResponse.put("healthy", result.getValue().isHealthy());
             resultResponse.put("message", result.getValue().getMessage());
+
+            if (result.getValue().getDetails() != null) {
+              for (Map.Entry<String, Object> entry : result.getValue().getDetails().entrySet()) {
+                resultResponse.put(entry.getKey().toString(), entry.getValue().toString());
```
alechenninger: nit: redundant toString on the key

Also, IMO let's add "timestamp" as a field, too, from getValue().getTimestamp()

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckRegistry.java
derek63: Not really, merged everything into CrudCheckRegistry

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckResponse.java
```diff
@@ -0,0 +1,41 @@
...
+        ObjectNode healthCheckResponse = factory.objectNode();
+        for (Map.Entry<String, HealthCheck.Result> result : healthCheckResults.entrySet()) {
+            ObjectNode resultResponse = factory.objectNode();
+            resultResponse.put("healthy", result.getValue().isHealthy());
+            resultResponse.put("message", result.getValue().getMessage());
+
+            if (result.getValue().getDetails() != null) {
+              for (Map.Entry<String, Object> entry : result.getValue().getDetails().entrySet()) {
+                  if(null != entry.getValue()) {
+                      resultResponse.put(entry.getKey().toString(), entry.getValue().toString());
```
alechenninger: nit: redundant toString() after getKey()

derek63: Fixed

crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckDiagnosticsCommand.java
```diff
@@ -66,18 +66,19 @@ public CallStatus run() {
                     String errorDetails = entry.getKey().toString() + " " + entry.getValue().getDetails();
                     Error.push("diagnosticCheck failed! " + errorDetails);
                     LOGGER.debug("diagnosticCheck failed!", errorDetails);
-                    return new CallStatus(Error.get(RestCrudConstants.ERR_REST_CHECK_DIAGNOSTICS, errorDetails));
+                    healthCheckResults.put(entry.getKey(), entry.getValue());
+                    healthCheckResponse = Response.Status.SERVICE_UNAVAILABLE;
```
alechenninger: Could probably simplify this a touch by putting healthCheckResults.put(entry.getKey(), entry.getValue()); outside of the conditional since it's in both branches.

crud/src/main/java/com/redhat/lightblue/rest/crud/health/ReadHealthCheck.java
```diff
@@ -0,0 +1,41 @@
...
+        String version = "1.0.0";
+        String request = getFindTestRequest(entity, version);
+
+        RequestMetrics metrics = new DropwizardRequestMetrics(MetricRegistryFactory.getJmxMetricRegistry());
+        CallStatus callStatus = new FindCommand(entity, version, request, metrics).run();
+
+        if(callStatus.hasErrors()){
+            return Result.unhealthy("Reads are not healthy: " + callStatus.getErrors());
+        }
+        return Result.healthy();
```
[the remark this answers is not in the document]
alechenninger: FWIW hasErrors was probably fine for the Find command, however I think this comment still needs to be addressed re: adding the details of the check to the result's details map.

crud/src/main/java/com/redhat/lightblue/rest/crud/cmd/CheckDiagnosticsCommand.java
alechenninger: Also, just double checking: so the verdict is all services' /diagnostics should return 503 when any check is unhealthy just like /health?

derek63: Yeah I briefly considered making diagnostics always return 200 (even if there is a failure), and health return 200 or 500 if it is working/failing, but was swayed by the following sentence in the 503 usage description:

> The implication is that this is a temporary condition which will be alleviated after some delay

So for consistency sake and to have one less mind map we have to maintain, both health and diagnostics will either return 200 when everything is passing or 503 when one or more things are failing.  If we see anything else (like a 500 or 404 or something), it is a problem with the deployment.  So whatever is checking these things (monitoring or load balancers) can rely on 200 response to indicate everything is okay, and anything else (503, 500, 404, etc) means there is a problem (the details of which should be reflected in the endpoint response).

derek63: > Looks good, just need the health check details added to the result and I think this is good to go.

This line 
> healthCheckResults.put(entry.getKey(), entry.getValue());
will add the details too since they are part of the HealthCheck.Result object, however I needed to add something in ReadHealthCheck to populate the details with the output of the test read (which is now done).

crud/src/main/java/com/redhat/lightblue/rest/crud/health/ReadHealthCheck.java
```diff
@@ -24,10 +24,16 @@ protected Result check() throws Exception {
...
-            return Result.unhealthy("Reads are not healthy: " + callStatus.getReturnValue());
+        ResultBuilder resultBuilder;
+
+        if (Response.Status.OK.equals(callStatus.getHttpStatus())) {
+            resultBuilder = Result.builder().healthy();
+        } else {
+            resultBuilder = Result.builder().unhealthy().withMessage("Reads are not healthy");
         }
-        return Result.healthy();
+        resultBuilder.withDetail("readResponse", callStatus.getReturnValue());
```
alechenninger: Does this serialize as expected? I expected you would have to use `callStatus.toJson()` which is what the regular responses uses (via toString() but point is RESTEasy isn't typically doing the marshalling for us so I'm not sure the call status bodies will all be serializable out of the box, I could be wrong tho). Possible to test?

derek63: So I did actually mean to use the toJson() here, so that was an oversight, but this also required a change in CrudCheckResponse.  The output now looks like this for healthy..

```
{
    "readHealthCheck": {
        "healthy": true,
        "message": null,
        "readResponse": {
            "entity": "test",
            "entityVersion": "1.0.0",
            "status": "COMPLETE",
            "modifiedCount": 0,
            "matchCount": 0,
            "hostname": "localhost",
            "processed": [],
            "resultMetadata": []
        }
    }
}
```
and like this for unhealthy
```
{
    "readHealthCheck": {
        "healthy": false,
        "message": "Reads are not healthy",
        "readResponse": {
            "entity": "test",
            "entityVersion": "1.0.0",
            "status": "ERROR",
            "modifiedCount": 0,
            "matchCount": 0,
            "hostname": "localhost",
            "errors": [
                {
                    "objectType": "error",
                    "context": "rest/FindCommand/test/find(test:1.0.0)/mongo:find",
                    "errorCode": "mongo-crud:ConnectionError",
                    "msg": "Prematurely reached end of stream"
                }
            ]
        }
    }
}
```

crud/src/main/java/com/redhat/lightblue/rest/crud/health/CrudCheckResponse.java
```diff
@@ -27,9 +27,13 @@ public JsonNode toJson() {
 
             if (result.getValue().getDetails() != null) {
               for (Map.Entry<String, Object> entry : result.getValue().getDetails().entrySet()) {
-                  if(null != entry.getValue()) {
-                      resultResponse.put(entry.getKey(), entry.getValue().toString());
+                if(null != entry.getValue()) {
+                  if(entry.getValue() instanceof ObjectNode) {
+                    resultResponse.put(entry.getKey(), (ObjectNode)entry.getValue());
+                  } else {
+                    resultResponse.put(entry.getKey(), entry.getValue().toString());
```
alechenninger: is just putting getValue() enough for all types and then we can let the serialization configuration figure out how to serialize?

derek63: Yes

PR #307, opened 2017-11-17
Add support for allAccessOu to login module

derek63:


derek63:
@snehagunta can you take a look when you have the chance?  This goes along with https://github.com/lightblue-platform/lightblue-puppet/pull/231

